PR #20, opened 2020-07-19
Email integration

cheeseblubber:
Using mailgun for sending emails


assets/src/components/demo/Demo.tsx
```diff
@@ -34,7 +34,7 @@ class Demo extends React.Component<Props, State> {
           title="Welcome to Papercups!"
           subtitle="Ask us anything in the chat window below 😊"
           primaryColor={colors.primary}
-          accountId="eb504736-0f20-4978-98ff-1a82ae60b266"
+          accountId="22af71fc-26b4-4c39-9425-e6e427be0b96"
```
reichert621, line 37: can we change this back? it's currently hardcoded to match our account in prod so that if anyone is trying out the demo page, we'll get pinged

(for testing locally i just update my account id to match this in my dev database)

README.md
```diff
@@ -14,6 +14,11 @@ To start your server:
 - Create and migrate your database with `mix ecto.setup`
 - Start the server with `mix phx.server`
 
+## Setting up email alerts
+
+- Set the environment variable in the .env.example
+* at the moment we only support amazon SES alerts
```
reichert621, line 20: mailgun for now instead, right?

.env.example
```diff
@@ -0,0 +1,6 @@
+export AWS_ACCESS_KEY_ID=YOUR_ACESS_KEY_ID
+export AWS_SECRET_ACCESS_KEY=YOUR_ACCESS_KEY
```
reichert621, line 2: should we remove env variables we're not using yet?

lib/chat_api_web/channels/conversation_channel.ex
```diff
@@ -38,6 +39,13 @@ defmodule ChatApiWeb.ConversationChannel do
   # broadcast to everyone in the current topic (conversation:lobby).
   @impl true
   def handle_in("shout", payload, socket) do
+    account = Accounts.get_account!(payload["account_id"])
+
+    account.users
+    |> Enum.filter(fn u -> u.email_alert_on_new_message end)
+    |> Enum.map(fn u -> u.email end)
+    |> Enum.map(fn e -> Email.send(e, payload["body"], payload["conversation_id"]) end)
```
reichert621, line 47: can we move all this logic to after line 56 where we create the message in the db? feels like we should create the message before sending the email. also there's no guarantee at this point that the `account_id` or `conversation_id` exist on the payload, so it would be safer to do it within the `case` statement below (though i'm realizing now that maybe a `with` clause would be more idiomatic? not sure...)

reichert621: (also might be nice to create a method to encapsulate `Emails.send_new_message_alerts(conversation_id, message)` or something rather than having all this logic here)

mix.exs
```diff
@@ -35,6 +35,9 @@ defmodule ChatApi.MixProject do
     [
       {:credo, "~> 1.4", only: [:dev, :test], runtime: false},
       {:dialyxir, "~> 1.0", only: [:dev], runtime: false},
+      {:hackney, "~> 1.9"},
```
reichert621, line 38: what's this dependency for btw?

cheeseblubber: Its a dependency required for Swoosh the email client it yells at us if we don't have it

assets/src/components/demo/Demo.tsx
cheeseblubber: Fixed

.env.example
cheeseblubber: fixed

lib/chat_api_web/channels/conversation_channel.ex
cheeseblubber: Moved it to Emails method to encapsulate it. I think we might have some sort of Alerter that takes a bunch of messaging channels and then picks which ones the user has turned on and sends them.

PR #32, opened 2020-07-24
Set up Conversations state manager for frontend

reichert621:
in-app-notifications

will definitely want to clean up this code over time... basically the problem we're trying to solve is:
- there are multiple "buckets" of conversations ("all", "assigned to me", "prioritized", "closed")
  - one slightly odd thing about these buckets is that there's some overlap... "all" will have every conversation that isn't closed (so everything in "assigned to me" and "prioritized" will also show up in "all"), and it's also possible that "assigned to me" and "prioritized" will have some overlap -- one partial solution could be to just distinguish between "assigned" and "unassigned", and then "prioritized" will not be its own bucket but just be a way of ordering conversations (maybe?)
- in the dashboard, i want to be alerted anytime there's a new message in an existing conversation, or a new conversation started
  - except if the conversation/message is in the "closed" bucket
- if i move a conversation from one bucket to another, i want to see a badge show up in the correct bucket (unless i closed it)
- if i'm viewing a conversation, i don't need to see a notification when a new message comes in
- if a conversation that's shared between two buckets is marked as "read" in one, it should also be marked in the other (i.e. the badge should decrement/disappear)
- (what else?)

how this code currently solves it:
- this PR moves all the conversation state to a single "React context", which makes it a bit easier to share state between components

the core of the state looks like this:
```js
{
  all: [1, 2, 3, 4], // list of conversation ids in the "all" bucket
  mine: [2, 3], // list of conversation ids in the "mine" bucket,
  closed: [5, 6, 7], // etc
  
  // map conversations by id as a single source of truth for each conversation
  conversationsById: {
    1: {status: 'open', read: true, ...etc},
    2: {status: 'open', read: false, ...etc},
    5: {status: 'closed', read: true, ...etc},
  },

  // likewise, map messages by conversation id for quick lookup
  messagesByConversation: {
    1: [{body: "hello world"}, {body: "sup"}, ...etc],
    2: [{body: "hi there"}, {body: "how can i help you"}, ...etc],
    5: [{body: "spam"}, {body: "test"}, ...etc]
  }
}
```

this makes it easier to get the conversations in each bucket, e.g.:
```js
const allConversations = all.map(id => conversationsById[id])
const currentMessages = messagesByConversation[selectedConversationId]
```

this is really similar to how state is often managed in redux, but i didn't want to introduce redux as a dependency because it's kind of an annoying thing to learn. this PR doesn't hide anything under the hood.

TODO: explain some of the edge cases and tricky parts and why we will need to continue refactoring this


assets/src/components/demo/Demo.tsx
```diff
@@ -29,7 +29,7 @@ class Demo extends React.Component<Props, State> {
       title: defaultTitle || 'Welcome to Papercups!',
       subtitle:
         defaultSubtitle || 'Ask us anything in the chat window below 😊',
-      accountId: 'eb504736-0f20-4978-98ff-1a82ae60b266',
+      accountId: 'fe038edd-ca6c-4072-a3dc-9350a773455d',
```
reichert621, line 32: revert

assets/src/components/demo/Demo.tsx
```diff
@@ -99,6 +99,7 @@ class Demo extends React.Component<Props, State> {
           subtitle={subtitle}
           primaryColor={color}
           accountId={accountId}
+          baseUrl="http://localhost:4000"
```
reichert621, line 102: revert

assets/src/components/conversations/ConversationsProvider.tsx
```diff
@@ -0,0 +1,519 @@
...
+    this.updateConversationState(conversations, 'closed');
+    this.handleJoinMultipleConversations(conversationIds);
+
+    return conversationIds;
+  };
+
+  getUnreadByCategory = () => {
+    const {all, mine, priority, conversationsById} = this.state;
+
+    // FIXME
```
reichert621, line 463: clarify

assets/src/components/conversations/AllConversations.tsx
```diff
@@ -1,63 +1,42 @@
...
-        .catch((err) => console.log('Error fetching account info:', err)),
-    ];
-
-    Promise.all(promises).then(() => this.setState({loading: false}));
+const AllConversations = () => {
+  const {
+    loading,
+    currentUser,
+    account,
+    // showGetStarted, // FIXME
```
reichert621, line 10: (i'll fix this before merging)

PR #37, opened 2020-07-26
Persist widget changes

cheeseblubber:
Just saving on component unmount for now give the user the option to save manually in the future
persist_widget_config


assets/src/api.ts
```diff
@@ -268,3 +275,19 @@ export const authorizeSlackIntegration = async (
...
+  widgetConfigParams: WidgetConfigParams,
+  id?: string,
+  token = getAccessToken()
+) => {
+  if (!token) {
+    throw new Error('Invalid token!');
+  }
+
+  return request
+    .put(`/api/widget_configs/${id}`)
```
reichert621, line 289: hmm i wonder if `widget_settings` would be a bit more "human readable"? (might just be me, but "configs" sounds kinda weird)

assets/src/api.ts
```diff
@@ -26,6 +26,13 @@ export type RegisterParams = LoginParams & {
   passwordConfirmation: string;
 };
 
+export type WidgetConfigParams = {
+  title: string;
+  subtitle: string;
+  color: string;
+  account_id: string;
```
reichert621, line 33: we don't need to pass this in explicitly since we can just get it from `conn.assigns.current_user.account_id` in the controller right?

assets/src/components/account/GettingStartedOverview.tsx
```diff
@@ -2,32 +2,37 @@ import React from 'react';
...
 };
 
 class GettingStartedOverview extends React.Component<Props, State> {
   state: State = {
-    account: null,
+    account_id: null,
     color: colors.primary,
     title: 'Welcome to Papercups!',
     subtitle: 'Ask us anything in the chat window below 😊',
+    // company_name: null,
```
reichert621, line 26: remove?

assets/src/components/account/GettingStartedOverview.tsx
```diff
@@ -2,32 +2,37 @@ import React from 'react';
...
   color: string;
   title: string;
   subtitle: string;
+  id?: string;
 };
 
 class GettingStartedOverview extends React.Component<Props, State> {
   state: State = {
-    account: null,
+    account_id: null,
```
reichert621, line 22: just a nitpick -- i know it's a little annoying to switch back and forth, but in general JS/TS code is camelCase rather than snake_case

assets/src/components/account/GettingStartedOverview.tsx
```diff
@@ -42,11 +47,25 @@ class GettingStartedOverview extends React.Component<Props, State> {
     this.setState({color: color.hex});
   };
 
+  async componentWillUnmount() {
```
reichert621, line 50: rather than handling this on unmount, it might be better to just auto-save on the "blur" event (i.e. when a user clicks out of an input field)

reichert621: (one reason: i ~think this won't get called if the user closes the window)

assets/src/api.ts
reichert621: i also don't think we need to pass in the widget_config `id`, if there's only one per account we should just do a lookup by `account_id` based on the authorized user -- what do you think?

assets/src/components/account/GettingStartedOverview.tsx
```diff
@@ -42,11 +47,25 @@ class GettingStartedOverview extends React.Component<Props, State> {
...
+    this.createOrUpdateWidgetConfig();
+  }
+
+  createOrUpdateWidgetConfig = () => {
+    const {color, title, subtitle, account_id} = this.state;
+    API.createOrUpdateWidgetConfig(
+      {color, title, subtitle, account_id},
+      this.state.id
+    )
+      .then((res) => console.log(res))
```
reichert621, line 60: happy to leave these in for now for debugging, but maybe add some text to make it more obvious what is being logged (e.g. `console.log('Updated widget config:', res)`, etc)

assets/src/components/account/GettingStartedOverview.tsx
```diff
@@ -163,12 +180,14 @@ const ExamplePage = () => {
...
             subtitle={subtitle}
             primaryColor={color}
-            accountId={accountId}
+            accountId={account_id}
             baseUrl={'https://app.papercups.io'}
             defaultIsOpen
           />
         </Box>
 
+        {/* <Button onClick={this.createOrUpdateWidgetConfig}>Save</Button> */}
```
reichert621, line 189: let's remove if this isn't being used

lib/chat_api/accounts.ex
```diff
@@ -37,7 +37,7 @@ defmodule ChatApi.Accounts do
 
   """
   def get_account!(id) do
-    Account |> Repo.get!(id) |> Repo.preload(:users)
+    Account |> Repo.get!(id) |> Repo.preload(:users) |> Repo.preload(:widget_config)
```
reichert621, line 40: I ~think we can just do `Repo.preload([:users, :widget_config])` rather than calling `preload` twice

lib/chat_api_web/controllers/slack_controller.ex
```diff
@@ -6,14 +6,8 @@ defmodule ChatApiWeb.SlackController do
   action_fallback ChatApiWeb.FallbackController
 
   def oauth(conn, %{"code" => code}) do
-    IO.inspect("Code from Slack OAuth:")
```
reichert621, line 9: can we leave these in for now? very useful for debugging slack integrations (until we've had a few people set up slack successfully, it's helpful to have these logs)

lib/chat_api_web/controllers/widget_config_controller.ex
```diff
@@ -0,0 +1,50 @@
...
+
+  action_fallback ChatApiWeb.FallbackController
+
+  def index(conn, _params) do
+    widget_configs = WidgetConfigs.list_widget_configs()
+    render(conn, "index.json", widget_configs: widget_configs)
+  end
+
+  def create(conn, %{"widget_config" => widget_config_params}) do
+    #TOOD Account id shouldn't be passed in #173997813
```
reichert621, line 15: can we handle this in this PR? shouldn't be much work right?

lib/chat_api_web/controllers/widget_config_controller.ex
```diff
@@ -0,0 +1,50 @@
...
+    with {:ok, %WidgetConfig{} = widget_config} <- WidgetConfigs.create_widget_config(widget_config_params) do
+      conn
+      |> put_status(:created)
+      |> put_resp_header("location", Routes.widget_config_path(conn, :show, widget_config))
+      |> render("show.json", widget_config: widget_config)
+    end
+  end
+
+  @spec update(any, map) :: any
+  def createOrUpdate(conn, %{"id" => id, "widget_config" => widget_config_params}) do
```
reichert621, line 25: nit -- elixir uses snake_case (`create_or_update`)

assets/src/api.ts
reichert621: also a little confused after looking at the backend code below -- shouldn't this be calling the `create_or_update` endpoint on this line? https://github.com/papercups-io/papercups/pull/37/files#diff-d7c965d3c254345cb15f54194253be75R49

lib/chat_api/widget_configs.ex
```diff
@@ -0,0 +1,118 @@
...
+      %WidgetConfig{}
+
+      iex> get_widget_config!(456)
+      ** (Ecto.NoResultsError)
+
+  """
+  def get_widget_config!(id), do: Repo.get!(WidgetConfig, id)
+
+  def create_or_update(nil, params) do
+      create_widget_config(params)
```
reichert621, line 41: looks like the indentation is a tiny bit off -- mind running `mix format` just to double check all the formatting?

assets/src/components/account/GettingStartedOverview.tsx
```diff
@@ -2,32 +2,37 @@ import React from 'react';
...
     title: 'Welcome to Papercups!',
     subtitle: 'Ask us anything in the chat window below 😊',
+    // company_name: null,
   };
 
   async componentDidMount() {
     const account = await API.fetchAccountInfo();
 
-    this.setState({account});
+    const {color, title, subtitle, id} = account.widget_config;
```
reichert621, line 32: this seems to break if no widget_config exists
Screen Shot 2020-07-26 at 2 57 38 PM

lib/chat_api_web/router.ex
```diff
@@ -46,12 +46,14 @@ defmodule ChatApiWeb.Router do
...
     get("/slack/oauth", SlackController, :oauth)
     get("/slack/authorization", SlackController, :authorization)
+    put("/widget_configs/create_or_update", WidgetConfigController, :createOrUpdate)
 
     resources("/user_invitations", UserInvitationController, except: [:new, :edit])
     resources("/accounts", AccountController, only: [:update, :delete])
     resources("/messages", MessageController, except: [:new, :edit])
     resources("/conversations", ConversationController, except: [:new, :edit, :create])
     resources("/customers", CustomerController, except: [:new, :edit, :create])
+    resources "/widget_configs", WidgetConfigController, except: [:new, :edit]
```
reichert621, line 56: probably don't need to expose every endpoint for the resource here, since all we care about is `create_or_update` and `show` (i.e. `PUT /api/widget_config` with auth and `GET /api/widget_config?account_id=xxxxx`)

assets/src/components/account/GettingStartedOverview.tsx
reichert621: can we get rid of this now that the blur callbacks are setup?

assets/src/components/account/GettingStartedOverview.tsx
reichert621: ^

lib/chat_api/widget_settings.ex
```diff
@@ -0,0 +1,118 @@
...
+      ** (Ecto.NoResultsError)
+
+  """
+  def get_widget_setting!(id), do: Repo.get!(WidgetSetting, id)
+
+  def create_or_update(nil, params) do
+      create_widget_setting(params)
+  end
+
+  def create_or_update(id, params) do
```
reichert621, line 44: rather than looking up by id, we could just do a `find_by_account_id` (instead of `get_widget_setting!(id)`) -- otherwise it could be possible for someone to update the widget settings of a different account if they knew the id

lib/chat_api_web/router.ex
```diff
@@ -46,12 +46,14 @@ defmodule ChatApiWeb.Router do
...
     get("/slack/oauth", SlackController, :oauth)
     get("/slack/authorization", SlackController, :authorization)
+    put("/widget_settings/create_or_update", WidgetSettingController, :create_or_update)
 
     resources("/user_invitations", UserInvitationController, except: [:new, :edit])
     resources("/accounts", AccountController, only: [:update, :delete])
     resources("/messages", MessageController, except: [:new, :edit])
     resources("/conversations", ConversationController, except: [:new, :edit, :create])
     resources("/customers", CustomerController, except: [:new, :edit, :create])
+    resources "/widget_settings", WidgetSettingController, only: [:index, :delete]
```
reichert621, line 56: if we're not using any of these endpoints, we can probably just remove this (we should be doing this more in general, but might as well start here)

PR #39, opened 2020-07-27
Start improving socket security

reichert621:
This PR just focuses on improving websocket security for logged in users in the `notifications_channel`

### DIFF
--- a/assets/src/components/conversations/ConversationsProvider.tsx
+++ b/assets/src/components/conversations/ConversationsProvider.tsx
@@ -97,7 +97,7 @@ export class ConversationsProvider extends React.Component<Props, State> {
   channel: Channel | null = null;
 
   async componentDidMount() {
-    socket.connect();
+    socket.connect({token: API.getAccessToken()});
 
     const [currentUser, account, numTotalMessages] = await Promise.all([
       API.me(),
@@ -279,19 +279,13 @@ export class ConversationsProvider extends React.Component<Props, State> {
     conversationId: string,
     cb?: () => void
   ) => {
-    const {account, currentUser} = this.state;
-    const {id: accountId} = account;
-    const {id: userId} = currentUser;
-
     if (!this.channel || !message || message.trim().length === 0) {
       return;
     }
 
     this.channel.push('shout', {
       body: message,
-      user_id: userId,
       conversation_id: conversationId,
-      account_id: accountId,
       sender: 'agent', // TODO: remove?
     });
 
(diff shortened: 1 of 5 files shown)


assets/src/components/conversations/ConversationsProvider.tsx
```diff
@@ -289,9 +289,7 @@ export class ConversationsProvider extends React.Component<Props, State> {
 
     this.channel.push('shout', {
       body: message,
-      user_id: userId,
       conversation_id: conversationId,
-      account_id: accountId,
```
reichert621, line 294: (these values are no longer necessary since they can be inferred from the auth token)

lib/chat_api_web/channels/notification_channel.ex
```diff
@@ -5,24 +5,19 @@ defmodule ChatApiWeb.NotificationChannel do
   alias ChatApi.{Chat, Conversations}
 
   @impl true
-  def join("notification:lobby", payload, socket) do
-    if authorized?(payload) do
-      {:ok, socket}
+  def join("notification:" <> account_id, %{"ids" => ids}, socket) do
+    if authorized?(socket, account_id) do
```
reichert621, line 9: (only allow joining the channel if the authed user's account matches the account id)

lib/chat_api_web/channels/notification_channel.ex
```diff
@@ -51,21 +46,28 @@ defmodule ChatApiWeb.NotificationChannel do
...
+      result = ChatApiWeb.MessageView.render("message.json", message: message)
 
-    case result do
-      %{conversation_id: conversation_id} ->
-        topic = "conversation:" <> conversation_id
+      # TODO: write doc explaining difference between push, broadcast, etc.
+      push(socket, "shout", result)
 
-        ChatApiWeb.Endpoint.broadcast_from!(self(), topic, "shout", result)
+      case result do
```
cheeseblubber, line 58: Any reason we are using case for a single statement check instead of just an if else?

cheeseblubber: Or are there other states we haven't handled yet and should maybe adds some todos here?

assets/src/components/conversations/ConversationsProvider.tsx
reichert621: yup exactly

lib/chat_api_web/channels/notification_channel.ex
reichert621: hmm i think we probably don't even need the `case` tbh, i think the `result` should always have a `conversation_id` if it reaches this point... maybe we could just do 
```ex
topic = "conversation:" <> result.conversation_id

ChatApiWeb.Endpoint.broadcast_from!(self(), topic, "shout", result)
```
i'll update 👍

PR #43, opened 2020-07-27
Set up Google Analytics

reichert621:
...with an environment variable!


.env.example
```diff
@@ -7,3 +7,6 @@ export DOMAIN=your_domain_where_you_send_email_from
 export SLACK_BOT_ACCESS_TOKEN=xoxb-xxxx-xxxx-xxxxxxxx
 export PAPERCUPS_SLACK_CLIENT_ID='xxxx.xxxx'
 export PAPERCUPS_SLACK_CLIENT_SECRET='xxxxxxxx'
+
+# React app
+export REACT_APP_GOOGLE_ANALYTICS_ID='UA-123456789-1'
```
cheeseblubber, line 12: We shouldn't commit this right?

PR #47, opened 2020-07-28
Set up API to update customer metadata from chat widget

reichert621:
This PR just sets up the API to be consumed from the widget (in this PR https://github.com/papercups-io/chat-widget/pull/8)

### DIFF
--- a/lib/chat_api_web/controllers/customer_controller.ex
+++ b/lib/chat_api_web/controllers/customer_controller.ex
@@ -36,6 +36,14 @@ defmodule ChatApiWeb.CustomerController do
     end
   end
 
+  def update_metadata(conn, %{"id" => id, "metadata" => metadata}) do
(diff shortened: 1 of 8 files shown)


lib/chat_api_web/controllers/customer_controller.ex
```diff
@@ -36,6 +36,14 @@ defmodule ChatApiWeb.CustomerController do
     end
   end
 
+  def update_metadata(conn, %{"id" => id, "metadata" => metadata}) do
+    customer = Customers.get_customer!(id)
```
cheeseblubber, line 40: Should we check that it belongs to the same account for these? Maybe we should have a helper function to check that the logged in user is the same one updating their resources

reichert621: ah this endpoint is public to non-logged in users since we need to send this data over from the widget

PR #49, opened 2020-07-29
Fix heroku deploy

cheeseblubber:


lib/chat_api/emails/email.ex
```diff
@@ -3,6 +3,7 @@ defmodule ChatApi.Emails.Email do
   import Ecto.Changeset
 
   @from_address System.get_env("FROM_ADDRESS")
+  @backend_url System.get_env("BACKEND_URL") || ""
```
reichert621, line 6: not necessary for this PR, but might be nice to set up a default value for these environment variables (i.e. localhost when in dev environment, app.papercups.io when in prod if no env variable is set)

PR #50, opened 2020-07-29
Add "online" status to customers in the dashboard

reichert621:
Handles #55

This PR uses Phoenix Presence to track when customers are online:
Screen Shot 2020-08-03 at 6 55 39 PM

### DIFF
--- a/assets/src/components/conversations/ConversationsProvider.tsx
+++ b/assets/src/components/conversations/ConversationsProvider.tsx
@@ -19,6 +19,7 @@ export const ConversationsContext = React.createContext<{
   unreadByCategory: any;
   conversationsById: {[key: string]: any};
   messagesByConversation: {[key: string]: any};
+  currentlyOnline: {[key: string]: any};
 
   onSelectConversation: (id: string | null) => any;
   onUpdateConversation: (id: string, params: any) => Promise<any>;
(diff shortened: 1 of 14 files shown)


assets/src/components/conversations/ConversationsProvider.tsx
```diff
@@ -168,6 +184,69 @@ export class ConversationsProvider extends React.Component<Props, State> {
       });
   };
 
+  handlePresenceState = (state: any) => {
+    this.setState({presence: state});
+  };
+
+  updatePresenceWithJoins = (joins: any, presence: any) => {
```
reichert621, line 191: (TODO: add types)

assets/src/components/conversations/ConversationsProvider.tsx
```diff
@@ -168,6 +184,69 @@ export class ConversationsProvider extends React.Component<Props, State> {
...
+    return Object.keys(joins).reduce((acc, key) => {
```
cheeseblubber, line 192: This return is kinda hard to understand. I m not really sure whats going on here. Also unclear on what is metas. Can we add some comments or maybe change the variable metas.

cheeseblubber: Maybe instead of calling them joins maybe called the NewCustomers or some sort of noun. having a verb as a variable is throwing me off a bit.

reichert621: sounds good, i'll add some comments to clarify and also add the type signatures 👍

PR #62, opened 2020-07-31
run with a docker container

rberrelleza:
Initial version of the Dockerfile.  It's not optimized for production just yet, but wanted to send something that works and then build up from there. 

Long term, I think moving to using [phoenix releases](https://hexdocs.pm/phoenix/releases.html) makes sense, since it allows you to have a much smaller docker image.  Also, not sure what you are using for releases, but it would be cool to hook this up for that (we use github releases to ship containers with well-known versions, it works pretty well)

Signed-off-by: Ramiro Berrelleza <rberrelleza@gmail.com>

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -1,11 +1,11 @@
 use Mix.Config
 
+database_url =
+  System.get_env("DATABASE_URL") || "ecto://postgres:postgres@localhost/chat_api_dev"
+
 # Configure your database
 config :chat_api, ChatApi.Repo,
-  username: "postgres",
-  password: "postgres",
-  database: "chat_api_dev",
-  hostname: "localhost",
+  url: database_url,
   show_sensitive_data_on_connection_error: true,
   pool_size: 10
 
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+set -euo pipefail
+
+POOL_SIZE=2 mix ecto.setup
+mix phx.server
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,38 @@
+# The directory Mix will write compiled artifacts to.
+/_build/
+
+# If you run "mix test --cover", coverage assets end up here.
+/cover/
+
+# The directory Mix downloads your dependencies sources to.
+/deps/
+
+# Where 3rd-party dependencies like ExDoc output generated docs.
+/doc/
+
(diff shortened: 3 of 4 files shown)


rberrelleza:
I also build a helm chart (so I could  run this in kubernetes), let me know if you are interested and I can send a separate PR.

config/dev.exs
```diff
@@ -1,11 +1,11 @@
 use Mix.Config
 
+database_url =
+  System.get_env("DATABASE_URL") || "ecto://postgres:postgres@localhost/chat_api_dev"
```
rberrelleza, line 4: I made this change so it was easier to test when running with non-local databases (e.g. in my case I run everything on a remote cluster, so i needed to change the hostname while still running in dev mode)

docker-entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -euo pipefail
+
+POOL_SIZE=2 mix ecto.setup
```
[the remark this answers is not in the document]
reichert621: i believe `ecto.setup` encapsulates both `ecto.create` and `ecto.migrate`, so i ~think it shouldn't be necessary

config/dev.exs
reichert621: nice!

reichert621:
thanks @rberrelleza! this is great 🎉 

cc @cheeseblubber mind taking this? you probably have more familiarity with this part of the code 😬

config/dev.exs
```diff
@@ -1,11 +1,11 @@
...
+
 # Configure your database
 config :chat_api, ChatApi.Repo,
-  username: "postgres",
-  password: "postgres",
-  database: "chat_api_dev",
-  hostname: "localhost",
```
reichert621, line 8: @cheeseblubber will this break things for us?

cheeseblubber: This is fine because its specified in the database URL

syabro:
@rberrelleza what's the difference between "dev" and "production" here?
I mean what's wrong if I delete node part from dockerfile and run the server?

rberrelleza:
@syabro the 'dev' and 'prod' env var controls how assets are build. You can't remove the node part, because you need it to build the webpack assets. 

There's a pending follow up on this task to optimize the docker image to make it smaller, so that the final image just have the assets. But the Dockerfile, as it is today, doesn't support that.

syabro:
@rberrelleza emm, actually I want to import react stuff in my main react application and use it there and keep the image clean

PR #65, opened 2020-07-31
Fix minor slack issue, add tests for slack controller

reichert621:


lib/chat_api_web/controllers/slack_controller.ex
```diff
@@ -111,12 +111,10 @@ defmodule ChatApiWeb.SlackController do
...
            event
        ) do
-    IO.inspect("Handling Slack event:")
-    IO.inspect(event)
-
-    thread = SlackConversationThreads.get_by_slack_thread_ts(thread_ts, channel)
+    # TODO: switch to `debug`
+    Logger.info("Handling Slack event: #{inspect(event)}")
 
-    with conversation <- thread.conversation do
```
reichert621, line 119: (this was the line that was breaking if `thread` was `nil`)

Issue #63, opened 2020-07-31
Support markdown in chat messages


reichert621:
Maybe give this a shot? https://marked.js.org/#/README.md

This could also work: https://github.com/rexxars/react-markdown

First goal would just be:
- If I type `**Hello** _world_ www.papercups.io`
- I see: **Hello** _world_ www.papercups.io

PR #77, opened 2020-08-04
Support markdown in chat messages

ZmagoD:
related to [issue]( https://github.com/papercups-io/papercups/issues/63)

Here I added the a component for the message body which sanitises and generates html from markdown.

Screenshot 2020-08-04 at 22 17 28
Screenshot 2020-08-04 at 22 15 53

TODO: Replicate the same logic in [chat-widget](https://github.com/papercups-io/chat-widget/) when the PR is approved


reichert621:
hey @ZmagoD thanks for tackling this! can we remove the margin from the bottom of the messages?

I'm not a fan of all this extra space at the bottom:
markdown-bug

reichert621:
also side note: if you ever want to chat with us directly about whatever you're working on, feel free to join our slack or discord channels https://github.com/papercups-io/papercups#get-in-touch 😄

ZmagoD:
@reichert621 I managed to get rid of the empty space now that I'm testing with [this](https://markdownlivepreview.com/) I see that the styles are a bit off. Do you have some style guides how it should look like?

Screenshot 2020-08-05 at 21 29 21
Screenshot 2020-08-05 at 21 29 26
Screenshot 2020-08-05 at 21 29 30

ZmagoD:
Another question 😅 

Should we permit to add images? The problem happens if someone inputs something like `/asset/images/sample.png`, we will get a server error.
Screenshot 2020-08-05 at 21 38 59

reichert621:
@ZmagoD we should probably not worry about images for now, does that error only happen when someone specifies a relative path? i'm not quite sure i understand how to reproduce 😅 

regarding general style, we're trying to stick as close to this as possible https://ant.design/components/typography/

(I'm tempted to experiment with https://github.com/rexxars/react-markdown because it might make it easier to customize how things are rendered by using react components)

sorry for all the thrash on this 😭

ZmagoD:
No problem @reichert621  😄 
Yes it makes more sense to use an already made component 😄 I will change the library, but maybe we will have the same problem with styles 🤔

reichert621:
@ZmagoD what are the main problems with the styles? is it mainly about keeping them the right color?

sorry this turned out to be trickier than i expected 😢 😛

reichert621:
I think the markdown we want to prioritize is probably just things like **bold**, _italics_, [links](www.reddit.com), maybe `this`, and whatever you think most people would need in messages like this -- headers tags are probably not important at the moment!

ZmagoD:
No problem @reichert621 😄 

yes basically the styles can be inconsistent because of the theme we use. I will add just bold italic links and pre. 🤔 maybe it will be better to use something like https://github.com/yabwe/medium-editor or http://wangzuo.github.io/react-medium-editor/ like slack uses WDYT?

reichert621:
hmm do those libraries support rendering the markdown when the message is sent, or just showing formatted text in the editor when typing? (it looks like it's only meant for the editor?) looks cool though!

ZmagoD:
I had a case like this in the past.
This library outputs HTML and I converted to markdown before save in the DB. When I rendered on the FE then I used https://github.com/rexxars/react-markdown to render it. But maybe this adds too much complexity in the project 🤔 
I'm not sure what would your users like more 🤔

reichert621:
@ZmagoD let's start off with just rendering the markdown correctly, and then handle improving the editor in another ticket 😛 but i really like the idea!

ZmagoD:
Screenshot 2020-08-08 at 15 23 06
Screenshot 2020-08-08 at 15 16 31

@reichert621 I changed the library and updated the styles

reichert621:
@ZmagoD this looks great!! thanks so much for doing this :) I'm gonna play around with it first and then will merge if it all works as expected 👍

Issue #59, opened 2020-07-31
[Low priority] Prevent users from signing up with an expired invitation


reichert621:
The `user_invitation` table has an `expires_at` field which should be taken into account when signing up. If the invite has expired, throw an error and let the user know.

reichert621:
handled in https://github.com/papercups-io/papercups/pull/81

Issue #58, opened 2020-07-31
Delete user invitation links after they are used


reichert621:
Right now user invitations (from `user_invitations` table) never get deleted, but we should delete them after they are successfully used.

This is where we currently use it: https://github.com/papercups-io/papercups/blob/master/lib/chat_api_web/controllers/registration_controller.ex#L14

If the registration is successful, the `user_invite` record should be deleted.

n1arash:
Hi,
I want to send a PR on this.
Isn't this better that we add a boolean flag for user invite and keep the record that will invalidate the invitation link and save the record in the database for future use like analytics or showing past invitations?

reichert621:
hmm yeah @n1arash, good point -- maybe we can add a flag like you suggested or just update the `expires_at` field to the current time, and then make sure we check against expiration before allowing the invitation to be used? what do you think?

n1arash:
I think updating the `expires_at` field can do the job, and also we can send an appropriate error to the frontend if a user used the link again with a 403 status header.

reichert621:
sounds perfect @n1arash

rlanga:
Hi,
I've done some sample modifications that delete the invite token after user creation is successful. I also split out the logic for cases when an invite token is included in the params and when it isn't. I think logic for #59 can be added to the create method for invite tokens. It could be modified to set the boolean flag to the invites as @n1arash mentioned.

Here's the [link](https://github.com/rlanga/papercups/commit/2632fdffb3e5095b819be698f7f052dd7c2173dc) to the commit on my repo fork.

reichert621:
@rlanga nice! your commit definitely looks like it solves the original issue, and i love the refactoring + test coverage 😁 

i think what @n1arash mentioned might be a good idea though — instead of deleting the invite after it's used, it would be helpful to keep it around so we can have a record of which invites have been used vs not. so rather than deleting, we could set up a method like
```elixir
ChatApi.UserInvitations.mark_user_invitation_expired(invite)
```

or something like that, and then throw a helpful error if someone tries to sign up with an expired invite. what do you think?

rlanga:
Definitely sounds like a great idea! I could modify my code for this

reichert621:
thanks @rlanga!

reichert621:
btw @rlanga and @n1arash if you ever want to chat with us directly about this, feel free to join our slack or discord channels https://github.com/papercups-io/papercups#get-in-touch 😄

reichert621:
[...]

PR #81, opened 2020-08-05
Processing user invitations on sign up

rlanga:
The code in this request deals with #58 and #59.

### DIFF
--- a/lib/chat_api_web/controllers/registration_controller.ex
+++ b/lib/chat_api_web/controllers/registration_controller.ex
@@ -16,7 +16,8 @@ defmodule ChatApiWeb.RegistrationController do
       |> Pow.Plug.create_user(params)
       |> case do
            {:ok, _user, conn} ->
-             ChatApi.UserInvitations.delete_user_invitation(invite)
+             ChatApi.UserInvitations.update_user_invitation(invite,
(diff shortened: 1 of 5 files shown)


lib/chat_api_web/controllers/registration_controller.ex
```diff
@@ -6,38 +6,67 @@ defmodule ChatApiWeb.RegistrationController do
...
+  def create(conn, %{"user" => user_params}) when is_map_key(user_params, "invite_token") do
+    try do
+      invite = ChatApi.UserInvitations.get_user_invitation!(user_params["invite_token"])
+      params = Enum.into(user_params, %{"account_id" => invite.account.id})
+      conn
+      |> Pow.Plug.create_user(params)
+      |> case do
+           {:ok, _user, conn} ->
+             if ChatApi.UserInvitations.expired?(invite) do
+               send_server_error(conn, "Invitation token has expired")
```
reichert621, line 20: hmm I wonder if we should use a 403 status for these errors instead of 500... not blocking for this PR, but just a note for the future

rlanga: Sure thing! That's a more informative error code

PR #82, opened 2020-08-05
Setup instructions via asdf-vm

jc00ke:
Hi there!

I've had a lot of success using `asdf` to manage language runtimes, and I think it offers the cleanest approach to keeping every developer in sync. Here's a quick attempt at changing the setup instructions; I'd love for someone who doesn't have `asdf` installed to run through them for QA.

Once change I would personally make, but hesitated to here, is to use a [per-project Postgres](https://jamey.thesharps.us/2019/05/29/per-project-postgres/) setup via [direnv](https://direnv.net/). I like the per-project setup, but it would be a slightly larger change that should go in another PR.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,0 +1,4 @@
+erlang 23.0.3
+elixir 1.10.4-otp-23
+nodejs 14.7.0
+postgres 12.3
--- a/README.md
+++ b/README.md
@@ -21,52 +21,23 @@ Stay tuned! We'll be posting a more detailed roadmap soon 🤓
 
 Papercups runs on Elixir/Phoenix, with a TypeScript React app for the frontend.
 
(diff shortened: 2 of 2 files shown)


reichert621:
very cool @jc00ke, thanks for doing this! i've never used asdf before, but will definitely give it a shot today :)

reichert621:
small update — after our little product hunt launch today we noticed a few more bugs, so we're trying to fix all those first... unfortunately didn't get around to trying this out, but i'm still excited to give it a shot soon!

(cc @cheeseblubber in case you have time to check this out before i do)

reichert621:
finally added this to our wiki! https://github.com/papercups-io/papercups/wiki/Development-Setup#setup-with-asdf-recommended

thanks again for the recommendation, and sorry for the delay 😅 had a lot of other things to worry about for a while!

Issue #113, opened 2020-08-10
Add ability to delete a conversation


fabborke:
For privacy reasons there should be the possibility to delete a conversation from the server.

[Edit from @reichert621]
Before we can fully support this, we'll need to fix some foreign key issues... namely, when a conversation is deleted, we should probably auto-delete all associated messages/slack threads/etc

reichert621:
great idea 👍

reichert621:
(need to QA this before closing)

reichert621:
(there were some issues with the PR above, need to fix some data issues before enabling this... e.g. db does not allow deleting certain resources due to current foreign key constraints)

reichert621:
done with https://github.com/papercups-io/papercups/pull/331 🎉

PR #145, opened 2020-08-14
add delete conversation button

ZmagoD:
Hi @reichert621 

this PR is related to issue https://github.com/papercups-io/papercups/issues/113 I added a button on the closed conversations view where the user can delete the conversation.

Screenshot 2020-08-14 at 19 55 44

The button is visible only on closed conversations page

### DIFF
--- a/assets/src/components/conversations/ConversationHeader.tsx
+++ b/assets/src/components/conversations/ConversationHeader.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+import React, {Fragment} from 'react';
 import {Box, Flex} from 'theme-ui';
 import {colors, Button, Select, Text, Title, Tooltip} from '../common';
 import {
   CheckOutlined,
+  DeleteOutlined,
   StarOutlined,
   StarFilled,
   UploadOutlined,
@@ -17,6 +18,7 @@ const ConversationHeader = ({
   onRemovePriority,
   onCloseConversation,
   onReopenConversation,
+  onDeleteConversation,
 }: {
   conversation: any;
   users: Array<any>;
@@ -25,6 +27,7 @@ const ConversationHeader = ({
   onRemovePriority: (conversationId: string) => void;
   onCloseConversation: (conversationId: string) => void;
   onReopenConversation: (conversationId: string) => void;
+  onDeleteConversation: (conversationId: string) => void;
 }) => {
   if (!conversation) {
     // No point in showing the header if no conversation exists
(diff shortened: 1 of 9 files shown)


reichert621:
@ZmagoD awesome!!! thanks so much for handling this, looks great so far 😄 

one small thing -- can we update the spacing between the buttons at the top to be consistent? looks like the "reopen" button is too close to the "delete" button (should have the same spacing as the space between "prioritize" and "reopen")

(let me know if that doesn't make sense)

also since deleting is probably something we want to be a bit more careful about, I wonder if we should do something like have a "confirmation" bubble before deleting? e.g. https://ant.design/components/popconfirm/ 

what do you think?

assets/src/components/conversations/ConversationHeader.tsx
```diff
@@ -114,12 +117,20 @@ const ConversationHeader = ({
...
-                />
-              </Tooltip>
+              <Fragment>
+                <Tooltip title="Reopen conversation" placement="bottomRight">
+                  <Button
+                    icon={<UploadOutlined />}
+                    onClick={() => onReopenConversation(conversationId)}
+                  />
+                </Tooltip>
+                <Tooltip title="Delete conversation" placement="bottomRight">
```
reichert621, line 127: to fix the spacing issue i mentioned in another comment, probably just need to render this in another `<Box mx={1}>` that only gets shown when `status === 'closed'`

assets/src/components/conversations/ConversationsContainer.tsx
```diff
@@ -169,6 +170,24 @@ class ConversationsContainer extends React.Component<Props, State> {
...
+  handleDeleteConversation = async (conversationId: string) => {
+    await this.props.onDeleteConversation(conversationId);
+
+    notification.open({
+      message: 'Conversation deleted!',
+      duration: 2, // 2 seconds
+      description: (
+        <Box>
+          You can view this conversations once again{' '}
+          <a href="/conversations/all">here</a>.
```
reichert621, line 182: hmm is this true? if the conversation gets deleted, i don't think it should show up anywhere anymore, right?

ZmagoD: ahh sorry copy paste error

ZmagoD:
Thank you for the review @reichert621 🙏 

I updated the code accordingly to your suggestions.


Screenshot 2020-08-14 at 21 02 21

assets/src/components/conversations/ConversationHeader.tsx
```diff
@@ -112,23 +116,40 @@ const ConversationHeader = ({
...
+                <Tooltip title="Reopen conversation" placement="bottomRight">
+                  <Button
+                    icon={<UploadOutlined />}
+                    onClick={() => onReopenConversation(conversationId)}
+                  />
+                </Tooltip>
+              </Box>
+              <Box mx={1}>
+                <Popconfirm
+                  title="Are you sure"
```
reichert621, line 131: can we change this to "Are you sure you want to delete this conversation?"

just want to make it as clear as possible what's going on :)

reichert621:
@ZmagoD just one last small thing, other than that looks great!

ZmagoD:
@reichert621 I updated the text content

Screenshot 2020-08-14 at 21 31 10

reichert621:
love it! 🎉

Issue #144, opened 2020-08-14
Save widget settings onChange (with debounce) instead of onBlur


reichert621:
If someone changes an input and then refreshes the page, the latest value won't persist because the "blur" event wasn't fired.

Let's handle updating the widget settings `onChange` (with maybe a 400ms debounce) instead, so the save happens automatically more or less.

All the changes will be in this file: https://github.com/papercups-io/papercups/blob/master/assets/src/components/account/GettingStartedOverview.tsx#L245

This is the page where these inputs live:
<img width="1359" alt="Screen Shot 2020-08-14 at 9 56 08 AM" src="https://user-images.githubusercontent.com/5264279/90257185-ce4bae80-de14-11ea-8d91-d02de25a2ea0.png">

reichert621:
@oddadmix let me know if you wanna take this!

Oddadmix:
Please assign it and I will work on it.

Oddadmix:
@reichert621, PR created 
https://github.com/papercups-io/papercups/pull/147

Oddadmix:
@reichert621 I can take up more UI issues/tasks. Let me know

reichert621:
thanks @Oddadmix i'll cc you in any good UI issues that come up :)

PR #147, opened 2020-08-14
Updating widget settings on change with debounce of 400 ms

Oddadmix:
Fix for https://github.com/papercups-io/papercups/issues/144


assets/src/components/account/GettingStartedOverview.tsx
```diff
@@ -8,6 +8,7 @@ import * as API from '../../api';
 import {User} from '../../types';
 import {Paragraph, Input, colors, Text, Title} from '../common';
 import {BASE_URL} from '../../config';
+import * as _ from 'lodash';
```
reichert621, line 11: let's just `import {debounce} from 'lodash'` if that's the only function we're using :) 

(i think this will also help optimize the webpack build to not import all of lodash)

assets/src/components/account/GettingStartedOverview.tsx
```diff
@@ -63,26 +64,38 @@ class GettingStartedOverview extends React.Component<Props, State> {
     }
   }
 
+  debounceedUpdateWidgetSettings = _.debounce(
```
reichert621, line 67: small typo here: `debounceedUpdateWidgetSettings` -> `debouncedUpdateWidgetSettings`

Issue #149, opened 2020-08-15
Auto assign conversation to yourself when you respond to the message


cheeseblubber:
Should only happen on the first response

reichert621:
(almost done in https://github.com/papercups-io/papercups/pull/154 but hasn't been completed yet)

PR #154, opened 2020-08-16
Auto assign conversation to yourself when you respond in a conversation for the first time

serranoenrique:
Hi @reichert621 ✋ 

## Problem

Conversations are manually assigned to a user at the moment. It would be much better if a conversation was automatically assigned to the first user to reply back in the conversation.

## Proposed solution

1. New query that counts how many messages have been sent by a company user in a conversation.
2. When a company user replies back, the notification channel handles the message and stores it.
3. If the new query is executed in that moment, the result should be 1.
4. Then we know the conversation should be assigned to that company user.

Related to #149 by @cheeseblubber.

### DIFF
--- a/lib/chat_api/messages.ex
+++ b/lib/chat_api/messages.ex
@@ -31,6 +31,16 @@ defmodule ChatApi.Messages do
     Repo.one(query)
   end
 
+  def count_messages_sent_by_company_user_in_conversation(conversation_id) do
+    query =
+      from(m in Message,
+        where: m.conversation_id == ^conversation_id and not is_nil(m.user_id),
+        select: count("*")
+      )
+
+    Repo.one(query)
+  end
+
   @doc """
   Gets a single message.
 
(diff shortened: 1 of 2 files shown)


lib/chat_api/messages.ex
```diff
@@ -31,6 +31,16 @@ defmodule ChatApi.Messages do
     Repo.one(query)
   end
 
+  def count_messages_sent_by_company_user_in_conversation(conversation_id) do
```
reichert621, line 34: can we add some tests for this method in `/test/chat_api/messages_test.exs`?

(I know we haven't been great about always adding tests, but I want to get more disciplined about it 😛 )

lib/chat_api_web/channels/notification_channel.ex
```diff
@@ -56,6 +58,14 @@ defmodule ChatApiWeb.NotificationChannel do
       message = Messages.get_message!(message.id)
       result = ChatApiWeb.MessageView.render("expanded.json", message: message)
 
+      # If the received message is the first one sent by any company user, assign the conversation to the user
+      if Messages.count_messages_sent_by_company_user_in_conversation(message.conversation_id) ==
```
reichert621, line 62: hmm I'm tempted to just encapsulate this logic in a method like
```ex
Conversations.has_company_user_replied?(conversation_id)
```
which would basically just check if the result of the count query is `> 0`... not sure that `@messages_limit_to_autoassign` is necessary at the moment

what do you think?

reichert621:
(closing, using https://github.com/papercups-io/papercups/pull/400 instead)

Issue #158, opened 2020-08-17
Show current user's email in profile page


reichert621:
For now let's just show it, we don't want to make it editable until we have a better email verification flow.

This is the section we want to show it in:
<img width="948" alt="Screen Shot 2020-08-16 at 9 32 16 PM" src="https://user-images.githubusercontent.com/5264279/90349578-62ca3280-e008-11ea-9aea-794ff0026897.png">

PR #160, opened 2020-08-17
Show current user's email in profile page #158

khabiirk:
I tried adding the current user email

### DIFF
--- a/assets/src/components/account/UserProfile.tsx
+++ b/assets/src/components/account/UserProfile.tsx
@@ -8,6 +8,7 @@ type State = {
   fullName: string;
   displayName: string;
   profilePhotoUrl: string;
+  email: string;
   shouldEmailOnNewMessages: boolean;
   isLoading: boolean;
   isEditing: boolean;
@@ -20,6 +21,7 @@ class UserProfile extends React.Component<Props, State> {
     fullName: '',
     displayName: '',
     profilePhotoUrl: '',
+    email: '',
     shouldEmailOnNewMessages: false,
     isLoading: true,
     isEditing: false,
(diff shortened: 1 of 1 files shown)


assets/src/components/account/UserProfile.tsx
```diff
@@ -83,6 +88,9 @@ class UserProfile extends React.Component<Props, State> {
   handleChangeProfilePhotoUrl = (e: any) => {
     this.setState({profilePhotoUrl: e.target.value});
   };
+  handleChangeEmail = (e:any) => {
```
rosinghal, line 91: Can we do something like `handleChangeEmail = (e: ChangeEvent<HTMLInputElement>) => {`?

assets/src/components/account/UserProfile.tsx
```diff
@@ -189,6 +199,16 @@ class UserProfile extends React.Component<Props, State> {
...
             />
           </Box>
+          <Box mb={3} sx={{maxWidth: 480}}>
+          <label htmlFor="email">Email:</label>
+          <Input
+            id="email"
+            type="text"
+            value={email}
+            onChange={this.handleChangeEmail}
+
```
rosinghal, line 209:
```suggestion
            disabled={!isEditing}
```

assets/src/components/account/UserProfile.tsx
```diff
@@ -189,6 +199,16 @@ class UserProfile extends React.Component<Props, State> {
               disabled={!isEditing}
             />
           </Box>
+          <Box mb={3} sx={{maxWidth: 480}}>
```
rosinghal, line 202: Position is wrong as https://github.com/papercups-io/papercups/pull/160/files#diff-f310b3aeb13547d64314e5e0a19aa695R213 shows photo, let's move it after line 190.

reichert621: +1 that the position is wrong, also since we're not making the email field editable let's just put it at the top :)

assets/src/components/account/UserProfile.tsx
reichert621: (+1 that adding better types here would be nice)

reichert621:
@khabiirk one other quick note -- when you create a pull request, can you add a link to the issue and include a screenshot of the changes? thanks again 🎉

reichert621:
closing, main fix handled in https://github.com/papercups-io/papercups/pull/219

Issue #157, opened 2020-08-17
Add company name to Account Overview page


reichert621:
Show the company name on the account overview page and make it editable (see user profile for editable input examples)

Let's add it in this section:
<img width="945" alt="Screen Shot 2020-08-16 at 9 32 04 PM" src="https://user-images.githubusercontent.com/5264279/90349518-2991c280-e008-11ea-8191-21e02d310725.png">

lukewaring:
Hi, is this a time-sensitive issue, or is there anything else I should know? I worked on it a little today and have everything functioning except getting the updates to persist properly. I'm aiming to wrap it up this week but just wanted to make sure it wasn't something more pressing. Thanks!

reichert621:
hey @lukewaring! it's not _too_ urgent, but definitely something I'd like to get done by end of week if possible :) feel free to open a WIP pull request in the meantime if you want!

PR #170, opened 2020-08-18
Add company name to Account Overview page (#157)

lukewaring:
Adds company name to Account Overview page and makes it editable (#157).

Screen Shot 2020-08-18 at 2 10 36 PM

### DIFF
--- a/assets/src/components/account/AccountOverview.tsx
+++ b/assets/src/components/account/AccountOverview.tsx
@@ -18,6 +18,7 @@ import {BASE_URL} from '../../config';
 type Props = {};
 type State = {
   account: any;
+  companyName: string;
   currentUser: any;
   inviteUrl: string;
   isEditing: boolean;
@@ -28,19 +29,31 @@ class AccountOverview extends React.Component<Props, State> {
 
   state: State = {
     account: null,
+    companyName: '',
     currentUser: null,
     inviteUrl: '',
     isEditing: false,
   };
 
   async componentDidMount() {
-    const [account, currentUser] = await Promise.all([
-      API.fetchAccountInfo(),
-      API.me(),
-    ]);
+    // NB: this fetches the account data and also handles setting this.state.account and this.state.companyName
+    await this.fetchLatestAccountInfo();
 
-    this.setState({account, currentUser});
+    const currentUser = await API.me();
+    this.setState({currentUser});
   }
+
+  fetchLatestAccountInfo = async () => {
+    const account = await API.fetchAccountInfo();
+
+    if (account) {
+      this.setState({account});
+      this.setState({companyName: this.state.account.company_name});
+    } else {
+      // NB: this also handles resetting this value if the optimistic update fails
+      this.setState({companyName: ''});
+    }
+  };
 
   handleGenerateInviteUrl = async () => {
     const {id: token} = await API.generateUserInvitation();
@@ -119,12 +132,7 @@ class AccountOverview extends React.Component<Props, State> {
   };
 
   handleChangeCompanyName = (e: any) => {
-    this.setState({
-      account: {
-        ...this.state.account,
-        company_name: e.target.value,
-      },
-    });
+    this.setState({companyName: e.target.value});
   };
 
   handleStartEditing = () => {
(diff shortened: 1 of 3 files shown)


assets/src/components/account/AccountOverview.tsx
```diff
@@ -112,25 +118,90 @@ class AccountOverview extends React.Component<Props, State> {
...
+    this.setState({
+      account: {
+        ...this.state.account,
+        company_name: e.target.value,
+      },
+    });
+  };
+
+  handleStartEditing = () => {
+    this.setState({isEditing: true});
```
reichert621, line 131: I think since we want to distinguish between what's showing up in the UI vs what's actually saved in the DB, it might be good to just add a `companyName` field to `this.state`, and then do something like set `this.setState({companyName: this.state.account.company_name})` here or whenever we get the `account` on mount, and then do a reset whenever the user hits "cancel"

one reason for this is if I'm editing the company name and then I hit cancel, I want to make sure the company name that is displayed in the UI goes back to whatever is actually saved

(this is how we currently handle it in the user profile: https://github.com/papercups-io/papercups/blob/master/assets/src/components/account/UserProfile.tsx#L88 basically if the user hits "cancel", we fetch the profile info from the server again and reset it... I imagine it would be good to do something similar here)

what do you think?

lukewaring: Hi Alex, thank you for the feedback here! That all makes a lot of sense, and I’ve made the changes you suggested.

One note - I based my `fetchLatestAccountInfo` method on the `fetchLatestProfile` method from the user profile. I carried over the `else` branch and the related comment from the user profile. I’m familiar with optimistic vs. pessimistic updates, but I'm not quite clear on how these else branches are functioning.

I just wanted to flag that, and I’m happy to keep it or make any updates. If easier, I can jump on Slack to chat too.

Sorry for the long comment - wanted to get everything in one place for you :) Thanks!

**AccountOverview.tsx**
```
fetchLatestAccountInfo = async () => {
   const account = await API.fetchAccountInfo();
  
   if (account) {
     this.setState({account});
     this.setState({companyName: this.state.account.company_name});
   } else {
     // NB: this also handles resetting this value if the optimistic update fails
     this.setState({companyName: ''});
   }
 };
```

**UserProfile.tsx**
```
fetchLatestProfile = async () => {
    const profile = await API.fetchUserProfile();

    if (profile) {
      const {
        display_name: displayName,
        full_name: fullName,
        profile_photo_url: profilePhotoUrl,
      } = profile;

      this.setState({
        displayName,
        fullName,
        profilePhotoUrl,
      });
    } else {
      // NB: this also handles resetting these values if the optimistic update fails
      this.setState({
        displayName: '',
        fullName: '',
        profilePhotoUrl: '',
      });
    }
  };
```

assets/src/components/account/AccountOverview.tsx
```diff
@@ -18,24 +18,43 @@ import {BASE_URL} from '../../config';
...
+    this.setState({currentUser});
   }
 
+  fetchLatestAccountInfo = async () => {
+    const account = await API.fetchAccountInfo();
+
+    if (account) {
+      this.setState({account});
+      this.setState({companyName: this.state.account.company_name});
+    } else {
```
reichert621, line 52: we can get rid of this block since an account should always exist :) 

(the only reason we had to do this in the profile is because of some tech debt where some users don't have a profile set up yet, so if the profile failed to save we needed to reset all the values)

assets/src/components/account/AccountOverview.tsx
```diff
@@ -18,24 +18,43 @@ import {BASE_URL} from '../../config';
...
+    const currentUser = await API.me();
+    this.setState({currentUser});
   }
 
+  fetchLatestAccountInfo = async () => {
+    const account = await API.fetchAccountInfo();
+
+    if (account) {
+      this.setState({account});
+      this.setState({companyName: this.state.account.company_name});
```
reichert621, line 51: i think we can simplify this to something like
```js
const account = await API.fetchAccountInfo();
const {company_name: companyName} = account;

this.setState({account, companyName});
```

what do you think?

assets/src/components/account/AccountOverview.tsx
reichert621: ah yes, sorry for the confusion 😅  i tried to explain in the comments below but let me know if anything doesn't make sense!

lukewaring: No worries and thank you for taking the time to explain! Definitely all makes sense now. I think we should be good to go :)

reichert621:
thanks @lukewaring! 🎉

lukewaring:
Happy to help @reichert621! 😀

Issue #80, opened 2020-08-05
Add support for webhooks


gdandersson:
Support for some kind of (web)hooks, for example when a chat is initiated and perhaps email has been entered, initiate hook which might return a list of orders from that customer (visible to the agents only, of course).

Or whenever a message is sent/received, a hook could look at the message and provide additional data (a simple regexp on the message might detect order numbers (x characters long and only numbers) which could then provide the agent with a direct URL to the order in the (ecommerce) backend.

cheeseblubber:
Thanks for opening the issue! Which event would be the most useful to you? This way we can prioritize these and break them down to smaller chunks. Also is there a specific production use case you are interested in for your website?

gdandersson:
If email is required to start chat, then that. Customer clicks on chat icon, enters email and "chat" -> webhook is fired.

The use case we'd love to use is looking up orders in our ecommerce system, so that agents quicker can see the order(s) the customer most likely will ask about.

So webhook of https://ecommerce.com/webhooks/papercups?email=a@b.com returns
1234 https://ecommerce.com/backend/order?id=1234
2543 https://ecommerce.com/backend/order?id=2543
(which agent sees)

.. for example. Makes sense?

reichert621:
great idea @gdandersson! makes a lot of sense.

for this to be useful for you, it sounds like we would need to do the following:
- when a new conversation is started (or a customer is created/updated), send a webhook event with the customer data (e.g. email, etc) in the form of a `POST` request to your webhook endpoint
- your endpoint would pick up the event, and query the customer's order data
- what do we do with that data? 
  - one idea: we let you add that data to the `conversation` in the form of a tag or internal note, so agents can see it alongside the conversation. we do this by exposing an API endpoint that allows you to set this metadata/note/whatever programmatically
  - or, did you have something else in mind? just want to make sure i understand :)

gdandersson:
In my world (with XMPP), I assume the hook is async so when data is returned, it's added as a message/internal note which is only sent to the agent.

I don't think any API endpoints to be exposed for this feature. I'm thinking a HTTP request to a specified URL and whatever is returned, is added as this note/message.

Guessing it could be more advanced with the answer having to be in some json/xml/whatever format. I'm just looking for a simple way of adding links to order(s) that my agents are dealing with.

reichert621:
gotcha, so just setting up a webhook event to be fired off via HTTP request to a specified URL when a conversation is started will be good enough for now?

[...]

PR #185, opened 2020-08-22
Set up webhook event subscriptions

reichert621:
Should handle https://github.com/papercups-io/papercups/issues/80

TODOs:
- [x] Set up `event_subscriptions` table 
- [x] Add ability to set up webhook URLs in the "Integrations" UI
- [x] Webhook URLs will need to be verified before they can be used (i.e. send a random string to the URL and make sure it replies with the correct string, the same way Slack verifies webhook URLs)
- [x] On `message:created`, send event payload to the account's verified webhook URLs
- (`conversation:created`, and `conversation:updated` events will be handled later)

Screen Shot 2020-08-23 at 10 23 53 PM
Screen Shot 2020-08-23 at 10 24 00 PM

### DIFF
--- a/assets/src/components/integrations/IntegrationsOverview.tsx
+++ b/assets/src/components/integrations/IntegrationsOverview.tsx
@@ -3,10 +3,22 @@ import {RouteComponentProps} from 'react-router-dom';
 import dayjs from 'dayjs';
 import {Box, Flex} from 'theme-ui';
 import qs from 'query-string';
-import {colors, Button, Paragraph, Table, Tag, Text, Title} from '../common';
+import {
+  colors,
+  Button,
+  Input,
+  Modal,
+  Paragraph,
+  Table,
+  Tag,
+  Text,
+  Title,
+} from '../common';
+import {PlusOutlined} from '../icons';
 import Spinner from '../Spinner';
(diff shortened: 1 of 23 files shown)


assets/src/components/integrations/IntegrationsOverview.tsx
```diff
@@ -16,16 +29,295 @@ type IntegrationType = {
   icon: string;
 };
 
+type WebhookEventSubscription = {
+  webhook_url: string;
+  verified: boolean;
+  created_at?: string | null;
+};
+
+const IntegrationsTable = ({
```
reichert621, line 38: (move these components into separate files)

assets/src/components/integrations/IntegrationsOverview.tsx
```diff
@@ -16,16 +29,295 @@ type IntegrationType = {
...
+        >
+          Save
+        </Button>,
+      ]}
+    >
+      <Box>
+        <Paragraph>
+          <Text>
+            You can subscribe to be notified of events in Papercups (for
+            example, when a new message is created) at a URL of your choice.
```
reichert621, line 250: (should update the wiki soon with details about setting up and testing webhooks, and then add a link here)

Issue #180, opened 2020-08-22
[feature request] disable signup, instead create a user when init


ruruiye:
papercups is a great software! I am look into the code for a while now, and want to deploy a instance into our company as customer support. but there is some issue came for a standard alone deployment:

Since our system need to access via internet, which means the login page and signup page is also available via internet, but in our case, we don't want to allow signup, instead we just need one account, and then invite user. so we need an option to disable signup. (better if no signup UI)

one option is to provide an option to disable signup feature, but enable an init user when start up. setup the user email, company, password, etc via system environment, and then register the user automaticallly.

MuhammadMullah:
yeah, I think I agree. We should probably the admin invite other users. This also reminds me of another feature which is part of the admin's roles. let me create a ticket for that

reichert621:
thanks @MuhammadMullah!

reichert621:
I believe this is fixed now, but feel free to reopen if not!

Issue #127, opened 2020-08-12
Ability to Disable Registration


masonhensley:
Hello,

I couldn't find a way to disable registration after creating the initial account (also tried to look through the source/docs).
Is there a way to prevent anyone from creating a new org on my deployment?

If it hasn't been build/spec'd yet, here's a potential direction:

**As an** Admin
**I want** to restrict outside registration
**So that** my instance of papercups is secure

Options:
- environment variable like `REGISTRATION_DISABLED=true` that would default to false if not set (might be the quickest V1)
- switch in UI

Other considerations:
- how does an admin know once they set it the deployment that other instances have not been registered?

Like the direction & plan to use this 🤓

reichert621:
hey @masonhensley! sorry for the delay on this -- we'll probably start off with just an environment variable like you suggested, but long term we'd like to have a better UI for managing this kind of thing :) 

to check if other instances have been registered, you would probably have to query your db through heroku -- happy to help you figure that out if you need!

reichert621:
closing for now, feel free to reopen if you notice any issues with this!

PR #188, opened 2020-08-25
Add env variable to make it possible to disable registration

reichert621:
Initial fix for https://github.com/papercups-io/papercups/issues/127 and https://github.com/papercups-io/papercups/issues/180

Users who are self-hosting can disable registration by setting the following environment variable:
```
export PAPERCUPS_REGISTRATION_DISABLED=1
# or
export PAPERCUPS_REGISTRATION_DISABLED=true
```

in Heroku, this might look like:
```
heroku config:set PAPERCUPS_REGISTRATION_DISABLED=true
```

When this is set, users will see the following error if they try to register:
Screen Shot 2020-08-25 at 12 54 12 PM

cc @masonhensley @ruruiye sorry for the delay on this!

### DIFF
--- a/lib/chat_api_web/controllers/registration_controller.ex
+++ b/lib/chat_api_web/controllers/registration_controller.ex
@@ -37,16 +37,20 @@ defmodule ChatApiWeb.RegistrationController do
   end
 
   def create(conn, %{"user" => user_params}) do
-    conn
-    |> user_with_account_transaction(user_params)
-    |> ChatApi.Repo.transaction()
-    |> case do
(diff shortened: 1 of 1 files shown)


masonhensley:
Right on! Thanks - I think this will help y'all with adoption from security minded folks. 

It's on a mental checklist when checking out self hosted software. 💯

johncadengo:
question @reichert621 if you enable this, how can you create an admin? Is the use case to quickly create the admin, then flip this env variable and restart server?

reichert621:
@johncadengo yup exactly

PR #192, opened 2020-08-28
Email Show on User Profile Page

seveibar:
This project has been [raided](https://www.twitch.tv/opensourceraiders)! Works locally.


seveibar:
okay maybe it doesn't actually work. but we tried

PR #196, opened 2020-08-31
Migration script to set `seen_at` field for old messages

reichert621:
Just using a simple heuristic for now: basically if a customer has replied to a message, we mark all previous messages as "seen"

### DIFF
--- a/priv/repo/migrations/20200831011352_set_messages_seen_at.exs
+++ b/priv/repo/migrations/20200831011352_set_messages_seen_at.exs
@@ -1,0 +1,46 @@
+defmodule ChatApi.Repo.Migrations.SetMessagesSeenAt do
+  use Ecto.Migration
+
+  import Ecto.Query, warn: false
+
+  alias ChatApi.{Messages, Repo}
+  alias ChatApi.Conversations.Conversation
+  alias ChatApi.Messages.Message
+
+  def up do
+    Conversation
+    |> preload(:messages)
+    |> Repo.all()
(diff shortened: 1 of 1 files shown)


priv/repo/migrations/20200831011352_set_messages_seen_at.exs
```diff
@@ -0,0 +1,44 @@
+defmodule ChatApi.Repo.Migrations.SetMessagesSeenAt do
+  use Ecto.Migration
+
+  import Ecto.Query, warn: false
+
+  alias ChatApi.{Messages, Repo}
+  alias ChatApi.Conversations.Conversation
+  alias ChatApi.Messages.Message
+
+  def up do
```
henrymori, line 10: May want to look into `change/0` which manages both up and down migrations without having to explicitly writing them out in your migration: https://hexdocs.pm/ecto_sql/Ecto.Migration.html#module-change

reichert621: i typically use `change/0`, but in this case i thought it wouldn't work 🤔 i'll give it a shot though!

reichert621: (didn't work in this case -- would've worked if i had done this in the last migration where i added the `seen_at` field in the first place... oops 😅)

henrymori: No worries, makes sense! 😄

Issue #57, opened 2020-07-31
Fix how we save IP addresses of customers


reichert621:
Since we currently get very little info on the anonymous customers that interact with the chat widget, we save their IP addresses (so we can eventually do a location lookup)

At the moment, we use `conn.remote_ip` (https://github.com/papercups-io/papercups/blob/master/lib/chat_api_web/controllers/customer_controller.ex#L17), but this ~seems to only return private IP addresses.

Investigate this as a possible solution? https://www.cogini.com/blog/getting-the-client-public-ip-address-in-phoenix/

henrymori:
I'll take a stab at this if it's not taken yet!

reichert621:
yes please @henrymori! (and as an added bonus, if you know of any good APIs to get locations from IP addresses, let me know 😬)

henrymori:
> yes please @henrymori! (and as an added bonus, if you know of any good APIs to get locations from IP addresses, let me know 😬)

Sounds good, thanks! I have some ideas on reverse geocoding I can kick around once we get the IP addresses accurately captured (at that point, we should probably start a separate GH issue for tracking IP -> geocoded information mapping).

reichert621:
for sure, will definitely want to track that in a separate ticket 👍

PR #197, opened 2020-08-31
Issue #57 | Fix for saving the correct IP address of customers

henrymori:
#### Summary. 

Implement a custom Plug (`ChatApiWeb.IPAddressPlug`) in the request pipeline to extract the value of the  `x-forwarded-for` header (when present) when Phoenix is running behind a reverse proxy (like is the case with Cloudflare, Heroku etc).

This implementation supports both IPv4 and IPv6 IP addresses. I looked into the solution at https://www.cogini.com/blog/getting-the-client-public-ip-address-in-phoenix/ but refrained from using too much logic from there as it (1) wasn't the most idiomatic Elixir code and (2) only seemed to work with IPv4 addresses.

Given this Plug runs on every request, extra consideration was made for performance and security (i.e. guarding against possible ForkBomb attacks with large `x-forwarded-for` values).

Tests accompany this PR covering both "happy" and "unhappy" code paths. The implementation has been tested both locally and on a PR environment in Heroku:

image

#### Ticket Link. 

Fixes: https://github.com/papercups-io/papercups/issues/57

### DIFF
--- a/lib/chat_api_web/ip_address_plug.ex
+++ b/lib/chat_api_web/ip_address_plug.ex
@@ -1,0 +1,55 @@
+defmodule ChatApiWeb.IPAddressPlug do
+  @moduledoc """
+  A plug to change the `Plug.Conn`'s `remote_ip` to use the values provided by the x-forwarded-for header in the event that the header exists.
+  """
+  @behaviour Plug
+
+  def init(opts \\ []), do: opts
+
+  def call(conn, _opts) do
+    ip = get_ip(conn.req_headers, nil)
+
+    case ip do
+      nil -> conn
+      _ -> %{conn | remote_ip: ip}
+    end
+  end
+
+  @spec parse_ip([{String.t(), String.t()}]) :: String.t() | nil
+  defp get_ip([], result), do: result |> truncate_header_value |> clean_ip |> parse_ip
+
+  defp get_ip([{"x-forwarded-for", value} | tail], _result),
+    do: get_ip(tail, value)
+
+  defp get_ip([_ | tail], result), do: get_ip(tail, result)
+
+  @spec truncate_header_value(String.t()) :: String.t()
+  @doc """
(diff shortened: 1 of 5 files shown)


henrymori:
@reichert621 I'd be keen to hear your thoughts on this approach

lib/chat_api_web/ip_address_plug.ex
```diff
@@ -0,0 +1,55 @@
...
+    do: get_ip(tail, value)
+
+  defp get_ip([_ | tail], result), do: get_ip(tail, result)
+
+  @spec truncate_header_value(String.t()) :: String.t()
+  @doc """
+    Given this is a "hot path" in the request lifecycle (served on every request) and Cowboy has a default header value size of 4096 bytes,
+    we want to avoid the possibility of forkbomb attacks (https://en.wikipedia.org/wiki/Fork_bomb).
+    IPv6 addresses are a maximum of 45 characters in length (https://stackoverflow.com/a/7477384)
+    Allowing for a reasonable buffer, only process the first 50 bytes of the header value
```
reichert621, line 31: 👍 nice

reichert621:
thanks so much for the quick turnaround on this @henrymori, looks great! (one of your commits also reminded me that we need to sort out the setup with asdf this week 😅)

hope you'll contribute more in the future!

henrymori:
@reichert621 My pleasure! Was definitely fun to work on this PR :) I'd also be happy to help with `asdf`, CI setup and a proposal for geocoding (which I'm also happy to tackle!). I'll follow up with separate GH issues 👍

reichert621:
that would be amazing! we already had someone submit a PR for the `asdf` setup a few weeks ago: https://github.com/papercups-io/papercups/pull/82 feel free to review it! (we ended up moving our dev setup docs into our wiki, so I was planning on just copying over whatever I can from that PR into our wiki or something like that)

PR #208, opened 2020-09-04
Refactor billing module

reichert621:
Just some attempts to make the code a bit more idiomatic 🤷

### DIFF
--- a/lib/chat_api/billing.ex
+++ b/lib/chat_api/billing.ex
@@ -88,7 +88,7 @@ defmodule ChatApi.Billing do
          {:ok, items} <- get_stripe_price_ids_by_product(product),
          {:ok, subscription} <-
            Stripe.Subscription.create(%{
-             customer: account.create_subscription_plan,
+             customer: account.stripe_customer_id,
              items: items,
              trial_period_days: @trial_period_days
            }) do
@@ -97,6 +97,8 @@ defmodule ChatApi.Billing do
         stripe_product_id: product.id,
         subscription_plan: plan
       })
+    else
+      error -> error
     end
   end
 
@@ -116,6 +118,8 @@ defmodule ChatApi.Billing do
         stripe_product_id: product.id,
         subscription_plan: plan
       })
+    else
+      error -> error
     end
   end
 end
--- a/lib/chat_api/billing.ex
+++ b/lib/chat_api/billing.ex
@@ -49,12 +49,9 @@ defmodule ChatApi.Billing do
   def find_stripe_product_by_plan(plan) do
     case Stripe.Product.list(%{active: true}) do
       {:ok, %{data: products}} ->
-        found =
-          Enum.find(products, fn prod ->
-            prod.metadata["name"] == plan
-          end)
-
-        case found do
+        products
+        |> Enum.find(fn prod -> prod.metadata["name"] == plan end)
+        |> case do
           nil -> {:error, :not_found}
           product -> {:ok, product}
         end
--- a/lib/chat_api/billing.ex
+++ b/lib/chat_api/billing.ex
@@ -6,6 +6,8 @@ defmodule ChatApi.Billing do
   import Ecto.Query, warn: false
   alias ChatApi.{Accounts, Messages}
 
+  require Logger
+
   @trial_period_days 30
 
   @doc """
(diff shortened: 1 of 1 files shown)


BobbyMcWho:
So I haven't actually messed with this myself, but looking at Sentry's docs, it seems like you can setup Sentry [at a plug level](https://github.com/getsentry/sentry-elixir#setup-with-plug-and-phoenix) and it will automatically capture errors for you. That way instead of manually capturing errors you can just let it fail and know Sentry will take care of it. 

Combined with [breadcrumbs or context](https://github.com/getsentry/sentry-elixir#context-and-breadcrumbs) I believe you could avoid having to do with do else blocks for all of them. 

If anyone with a bit more experience using Sentry with elixir has a reason why this wouldn't work, please chime in 😅

BobbyMcWho:
I guess a concern I have with returning `nil` if any 1 call fails is that then you don't have a reliable set of data. One time you may get 
```elixir
  %{
      subscription: "some_subscription",
      product: nil,
      payment_method: "VISA",
      subscription_plan: "PREMIUM",
      num_messages: 1000,
      num_users: 5
    }
```
and the next time you call it you could get
```elixir
  %{
      subscription: nil,
      product: "PAPERCUPS",
      payment_method: "VISA",
      subscription_plan: nil,
      num_messages: 1000,
      num_users: 5
    }
```

Does your app properly function in both of those scenarios? 

If not, then you probably don't want to bypass the errors and return `nil`

reichert621:
thanks for the feedback @BobbyMcWho! 

if i just did something like
```ex
def retrieve_stripe_resource(_resource, nil), do: nil

def retrieve_stripe_resource(:subscription, subscription_id) do
  with {:ok, subscription} <- Stripe.Subscription.retrieve(subscription_id) do
    subscription
  end
end

def retrieve_stripe_resource(:product, product_id) do
  with {:ok, product} <- Stripe.Product.retrieve(product_id) do
    product
  end
end

# etc
```

do you know if this is the idiomatic way to basically just return the `result` portion of `{:ok, result}` as long as the id is not `nil`?

BobbyMcWho:
In that case, you wouldn't actually need the with, you can just plain pattern match
```elixir
{:ok, product} = Stripe.Product.retrieve(product_id)
product
```

But here I'd actually just do

```elixir
Stripe.Product.retrieve(product_id)
```

Then I'd do the `with` block up above when you call each function.

reichert621:
hmm yeah the only problem is `Stripe.Product.retrieve(product_id)` blows up if `product_id` is `nil`, but I just want to return `nil` in that case which is why I set up this line to handle that case: https://github.com/papercups-io/papercups/pull/208/files#diff-4d3739183f331805332bf600ca31ebf7R32

but it does feel a little silly... is there a better way to just have the `Stripe.retrieve` methods return nil if the argument is nil?

BobbyMcWho:
It'd be pretty similar to what you have actually, but instead of
```elixir
with subscription <- yadda
```
You'd do

```elixir
with {:ok, subscription} <- yadda
```

And handle _any_ error tuple in the else.

reichert621:
ah so basically something like this?
```ex
def get_billing_info(account) do
  with {:ok, subscription} <- retrieve_stripe_resource(:subscription, account.stripe_subscription_id),
       {:ok, product} <- retrieve_stripe_resource(:product, account.stripe_product_id) do
    %{
      subscription: subscription,
      product: product
      # etc
    }
  end
end

def retrieve_stripe_resource(_resource, nil), do: nil

def retrieve_stripe_resource(:subscription, subscription_id),
  do: Stripe.Subscription.retrieve(subscription_id)

def retrieve_stripe_resource(:product, product_id),
  do: Stripe.Product.retrieve(product_id)
```

BobbyMcWho:
Yup, but instead of 
```elixir
def retrieve_stripe_resource(_resource, nil), do: nil
```

```elixir
def retrieve_stripe_resource(_resource, nil), do: {:error, :nil_resource_id}
```

reichert621:
in my case, having a `nil` id is a valid state, which is why i just want to return `nil` instead of trying to retrieve the resource from stripe :)

BobbyMcWho:
> in my case, having a `nil` id is a valid state, which is why i just want to return `nil` instead of trying to retrieve the resource from stripe :) 

Ah okay, sounds good

reichert621:
ah yes, good catch! thanks again for help @BobbyMcWho 😄

PR #215, opened 2020-09-08
Setup customerio

cheeseblubber:


lib/chat_api_web/controllers/registration_controller.ex
```diff
@@ -80,6 +80,34 @@ defmodule ChatApiWeb.RegistrationController do
     conn
   end
 
+  defp send_registration_event(conn) do
+    with %{email: email, account_id: account_id} <- conn.assigns.current_user do
+      %{email: email}
```
reichert621, line 85: can we remove this line?

lib/chat_api_web/controllers/registration_controller.ex
```diff
@@ -80,6 +80,34 @@ defmodule ChatApiWeb.RegistrationController do
...
+
+      now = :os.system_time(:seconds)
+
+      param = %{
```
reichert621, line 89: nit - we can probably just inline this below if it's not used anywhere else 🤷

lib/chat_api_web/controllers/registration_controller.ex
```diff
@@ -80,6 +80,34 @@ defmodule ChatApiWeb.RegistrationController do
...
+        email: email,
+        created_at: now,
+      }
+
+
+      Customerio.identify(account_id, param)
+      |> case do
```
reichert621, line 96: nit - no need to pipe here:
```ex
case Customerio.identify(account_id, %{email: email, created_at: now}) do
  {:ok, _} -> nil
  {:error, result} -> Logger.error(result)
end
```

lib/chat_api_web/controllers/registration_controller.ex
```diff
@@ -80,6 +80,34 @@ defmodule ChatApiWeb.RegistrationController do
...
+        {:ok, _} -> nil
+        {:error , result} -> IO.inspect(result)
+      end
+
+      Customerio.track(account_id, "sign_up", %{signed_up_at: now})
+      |> case do
```
reichert621, line 102: (same as above)

lib/chat_api_web/controllers/registration_controller.ex
```diff
@@ -80,6 +80,34 @@ defmodule ChatApiWeb.RegistrationController do
...
+      param = %{
+        email: email,
+        created_at: now,
+      }
+
+
+      Customerio.identify(account_id, param)
+      |> case do
+        {:ok, _} -> nil
+        {:error , result} -> IO.inspect(result)
```
reichert621, line 98: let's also use `Logger.error` here instead of `IO.inspect` so errors will get logged to Sentry 👍 

(we'll need to `require Logger` at the top for this to work)

lib/chat_api_web/controllers/registration_controller.ex
```diff
@@ -23,7 +23,7 @@ defmodule ChatApiWeb.RegistrationController do
             # # TODO: figure out what we want to do here -- it's not currently
             # # obvious that a user invitation expires after one use.
             # ChatApi.UserInvitations.expire_user_invitation(invite)
-            conn |> enqueue_welcome_email() |> send_api_token()
+            conn |> send_registration_event |> enqueue_welcome_email() |> send_api_token()
```
reichert621, line 26: nit - let's use parens here for consistency (i think it's generally encouraged to use parens even when they can be omitted)

.env.example
```diff
@@ -26,3 +26,7 @@ export REACT_APP_URL='app.papercups.io'
 
 # General
 export BACKEND_URL='app.papercups.io'
+
+# Customer.io
+export CUSTOMER_IO_SITE_ID='0000000000000000000'
+export CUSTOMER_IO_API_KEY='xxxxxxxxxxxxx'
```
reichert621, line 32: nit - let's add a newline to the end of the file to make that red thing go away :P

reichert621 (commented):
woohoooo just a couple nits then feel free to merge

lib/chat_api_web/controllers/registration_controller.ex
```diff
@@ -80,6 +80,34 @@ defmodule ChatApiWeb.RegistrationController do
...
+
+      now = :os.system_time(:seconds)
+
+      param = %{
+        email: email,
+        created_at: now,
+      }
+
+
+      Customerio.identify(account_id, param)
```
reichert621, line 95: just curious, why account ID and not user ID here?

cheeseblubber: Good catch actually should be user_id

Issue #151, opened 2020-08-15
Continue over email


cheeseblubber:

timgl:
+1, this combined with #165 is crucial as otherwise we have no way of getting back to users when we're not online

reichert621:
I'm thinking for the first version of this, we just do something like:
- **if**
  - the agent responds, 
  - and the customer is no longer "online" (i.e. active on the website with the widget), 
  - and we have a valid email address available for the customer
- **then** 
  - we send the customer an email from `{agent.name} <hello@papercups.io>` with the latest N messages, 
  - with a `reply_to: {agent.email}` in case the customer wants to reply directly via email, 
  - and a link to the page they were at when they last interacted with the widget

(for now, we won't worry about tracking replies via email)

@timgl @rosinghal @cheeseblubber let me know if any of that doesn't make sense!

timgl:
Yep I think that works for me!

rosinghal:
@reichert621 can we make `reply_to` configurable? Reason is I can use some generic email like support@example.com as some generic email where other agents can take care of it as well

reichert621:
hey @rosinghal! yes, I think we should be able to do that -- would it be ok if that was an account-level setting? (so that no matter which agent sent the message, the reply_to would always be e.g. `support@example.com`?)

rosinghal:
@reichert621 yup makes sense.

reichert621:
notes from #107 
> Right now if a customer goes offline and we respond, they won't see the message unless they come back to the page and open the widget -- would be nicer if they could get answers to their questions/comments via email if possible.

> I also will like to see the possibility of continue conversations by email.
> I suggest SMTP as a base integration (as most providers have this option) then add integrations with other providers (my prefered provider is Mailgun as it has a nice API and EU/US servers)

reichert621:
@rosinghal @timgl this is almost done -- just testing in staging first! hoping to deploy later today if things look good 👍

reichert621:
this should be working now! let me know if you have any questions, but basically it will work like this: https://github.com/papercups-io/papercups/issues/151#issuecomment-688336245

(the only update is that we wait 2 minutes to send the email, to give the customer some buffer to see it before receiving an email, and also to give the agent time to send multiple messages without having multiple emails go out)

fullpwemium:
Unable to get this to work anymore....  self-hosting and with the env variable enabled I see the following:

"23:51:37.416 info Skipped sending Gmail notification: nil"

PR #216, opened 2020-09-09
Set up "conversation reply" email

reichert621:
Sets up email notifications when customer goes offline and agent is replying to a conversation.

(Will be disabled until the `CONVERSATION_REPLY_EMAILS_ENABLED` environment variable is turned on)

Will handle https://github.com/papercups-io/papercups/issues/151


lib/workers/send_conversation_reply_email.ex
```diff
@@ -0,0 +1,83 @@
+defmodule ChatApi.Workers.SendConversationReplyEmail do
+  use Oban.Worker,
+    queue: :mailers,
+    unique: [
```
reichert621, line 4: (still need to figure out the best way to handle this)

PR #226, opened 2020-09-14
Test out mnesia cache for auth persistence (WIP)

reichert621:
See https://hexdocs.pm/pow/README.html#pow-plug-session


reichert621:
closing, doesn't work with heroku unfortunately :( may need to use redis or something else instead

PR #227, opened 2020-09-14
Fix up docker-compose.yml

lingster:
include only docker fixes in this PR

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -2,4 +2,6 @@
 set -euo pipefail
 
 POOL_SIZE=2 mix ecto.setup
+mix deps.compile certifi
+echo "Run: mix phx.swagger.generate to generate swagger docs" 
 mix phx.server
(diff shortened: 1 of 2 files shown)


lingster:
@reichert621 : new PR as requested. Note that this will expose the postgres db container to port 5432, so you will be able to run papercups locally on your machine to debug and connect onto the postgres db in the docker container.

docker-entrypoint.sh
```diff
@@ -2,4 +2,6 @@
 set -euo pipefail
 
 POOL_SIZE=2 mix ecto.setup
-mix phx.server
\ No newline at end of file
+mix deps.compile certifi
+echo "Run: mix phx.swagger.generate to generate swagger docs"
```
reichert621, line 6: should we remove this for now?

reichert621 (commented):
thanks @lingster! just one comment :)

Issue #137, opened 2020-08-13
Support 2FA in auth flow


reichert621:
Let's make our auth flow better with 2 factor auth!

PR #235, opened 2020-09-17
Add "admin" vs "user" roles

reichert621:
This will allow us to restrict certain actions in the dashboard to admin users (e.g. inviting teammates, removing teammates)

This PR:
- [x] Adds the `role` field to the `users` table
- [x] Defaults all current users to `admin` role
- [x] Restricts user invitations to admin users
- [x] Updates the Account Overview UI to display user roles

(We'll handle allowing admin users to remove teammates in a future PR)

Screen Shot 2020-09-18 at 12 50 09 PM

### DIFF
--- a/assets/src/components/account/AccountOverview.tsx
+++ b/assets/src/components/account/AccountOverview.tsx
@@ -52,14 +52,18 @@ class AccountOverview extends React.Component<Props, State> {
   };
 
   handleGenerateInviteUrl = async () => {
-    const {id: token} = await API.generateUserInvitation();
+    try {
+      const {id: token} = await API.generateUserInvitation();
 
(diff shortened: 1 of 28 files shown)


masonhensley:
Nice @reichert621 - have you thought about role inheritance from SSO providers? Might be able to knock out #137 by delegating it to [Keycloak](https://github.com/keycloak/keycloak), gSuite, etc.

Might be able to knock out multiple birds with one stone, not sure how that fits onto your enterprise roadmap.

reichert621:
thanks @masonhensley, i'll definitely look into that! i honestly don't have much experience with e.g. keycloak, but it seems like a great idea 👍

PR #242, opened 2020-09-22
Allow admin users to disable members on their account

reichert621:
Screen Shot 2020-09-22 at 2 28 34 PM
Screen Shot 2020-09-22 at 2 26 00 PM

### DIFF
--- a/assets/src/api.ts
+++ b/assets/src/api.ts
@@ -559,3 +559,31 @@ export const createPaymentMethod = async (
     .set('Authorization', token)
     .then((res) => res.body.data);
 };
+
+export const disableAccountUser = async (
+  userId: number | string,
+  token = getAccessToken()
+) => {
+  if (!token) {
+    throw new Error('Invalid token!');
+  }
+
+  return request
+    .post(`/api/users/${userId}/disable`)
+    .set('Authorization', token)
+    .then((res) => res.body.data);
+};
+
+export const enableAccountUser = async (
+  userId: number | string,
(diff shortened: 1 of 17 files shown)


reichert621 (commented):
add some more tests and then this should be good to go!

MuhammadMullah:
I am not a react guy. But I have gone through the whole code and more specifically the elixir code and it looks good to me. But out of curiosity or let me say learning from each other. For me, I would have made disabled_at to be a boolean with default as false, and whenever I disable a user its just a matter of updating it to true.

maybe if you can shed light on me on the decision on why you made disabled_at as utc_datetime?

MuhammadMullah:
also couldn't some of those functions be private e.g archive_user, disable_user, enable_user, or are they being used anywhere else outside of that module?

reichert621:
hey @MuhammadMullah! great questions :) 

- in terms of using `disabled_at` as a datetime rather than a boolean, this is a common pattern used because it gives you more information with the same functionality -- for example, it's often helpful to know exactly when a user was archived or disabled, and you can infer if the user is disabled based on the presence of that field (i.e. if it's `null`, then they user is not disabled)
- those methods are currently used in the `UserController`, so they can't be private unfortunately 

let me know if you have anymore questions!

MuhammadMullah:
> * those methods are currently used in the `UserController`, so they can't be private unfortunately
>  ooh maybe I was not careful to check that. Okay great, thanks

Issue #85, opened 2020-08-05
[Low priority] Consider removing `create_or_update`s


reichert621:
Instead of doing a `create_or_update` for our user profile, user settings, widget settings, it might make more sense to just create these initial records when a user signs up for the first time and just do `updates` in the future. 

See this thread: https://github.com/papercups-io/papercups/pull/74#discussion_r465968811

Issue #60, opened 2020-07-31
Set up factories for our tests to DRY them up


reichert621:
See https://hexdocs.pm/ecto/test-factories.html

Currently a bunch of our tests have the same boilerplate code... let's try to DRY it up with factories!

MuhammadMullah:
@reichert621 I'd like to attempt getting this one done

reichert621:
@MuhammadMullah that sounds great! nothing you should need to get started, other than setting up your development environment: https://github.com/papercups-io/papercups/wiki/Development-Setup#getting-started

basically we have a lot of repeated boilerplate code in our tests and would like to clean it up with factories, so just make sure `mix test` still passes before opening a PR 😄

MuhammadMullah:
great

MuhammadMullah:
@reichert621 working on this isssue but I am a bit unsure whether this has been done already. Just want to confirm if I can still continue with this

reichert621:
hey @MuhammadMullah! a bunch of the fixtures were refactored already, but I still think it would be really useful to set up factories (https://hexdocs.pm/ecto/test-factories.html) for a bunch of our models (e.g. messages, conversations, accounts, etc.)

so there is still some work to do here if you want to pick it up!

SherSpock:
Hey @MuhammadMullah! Hope you're well. Just checking if you are still looking at this.

reichert621:
@SherSpock given how long it's been, I'm happy to assign this to you if you're interested!

SherSpock:
@reichert621 I am!

Forgive me if I'm overcomplicating this, but there's two things I wanted to get your input about before starting anything:

1) Could we split this work into multiple work items? 

2) Could we have a separate module for each factory? 

The way the Ecto documentation does factories is reminiscent of how factories are done with `factory_bot` and Laravel -- it lumps everything onto an all-knowing factory. This is nice in that you can find all your factories in one file, and less nice when your one file gets cluttered or you'd like Intellisense to tell you what factory definitions are available for a given model. And if in the all-knowing factory case, you start pulling your factory definitions into separate files to better organize them, I feel like you have now reached the worst of both worlds -- you have neither Intellisense (at least not without writing or using an extra plugin that understands these definitions) nor separation of concerns, but your file organization could easily lead an unsuspecting dev to expect both.

reichert621:
@SherSpock sure!

[...]

PR #246, opened 2020-09-24
Set default values for account & user upon creation

adipurnama:
Hi,

I was trying to implement #85 and while updating the tests, I thought it might also good idea to create fixture helper for the tests. While running the tests, i noticed it can take quite some time (20-40s). I change some of the tests file to enable  async test execution for faster `mix test`. 

What's changed:

* Set default values for user_settings, user_profile & widget_settings
   upon account signup & user creation. Change from `create_or_update` to
   `update` for controllers & contexts. (#85)

* Add test fixture helpers to DRY up test files. (#60)

* Async ExUnit test execution for faster `mix test`, except `channel` related tests, some of it fails when I enabled it. :(


reichert621:
thanks @adipurnama, this looks amazing! 🎉 

looks like a pretty big diff, so i'll review this by the end of the day :)

lib/chat_api/accounts.ex
```diff
@@ -53,9 +54,20 @@ defmodule ChatApi.Accounts do
...
-    %Account{}
-    |> Account.changeset(attrs)
-    |> Repo.insert()
+    result =
+      Ecto.Multi.new()
+      |> Ecto.Multi.insert(:account, Account.changeset(%Account{}, attrs))
+      |> Ecto.Multi.run(:widget_settings, fn _repo, %{account: account} ->
+        %WidgetSetting{}
+        |> WidgetSetting.changeset(%{account_id: account.id})
+        |> Repo.insert()
```
reichert621, line 63: just curious, why `Repo` here instead of the `_repo` above?

lib/chat_api/accounts.ex
```diff
@@ -88,8 +100,22 @@ defmodule ChatApi.Accounts do
...
-    Repo.delete(account)
+  def delete_account(%Account{id: id} = account) do
+    result =
+      Repo.transaction(fn ->
+        setting =
+          WidgetSetting
+          |> where(account_id: ^id)
+          |> Repo.one()
+
+        Repo.delete(setting)
```
reichert621, line 111: not necessary for this PR, but we should probably eventually just handle this at the db level 🤔

reichert621 (commented):
looks great so far @adipurnama -- i think we might also need to do a small migration for any users that have registered but never bothered to edit their profile info (in which case a `user_profile` record never would've been created)

lib/chat_api/accounts.ex
adipurnama: You're right. Using the previous `repo` variable might be better for `Ecto.Multi.run` block.
I just copy-pasted it from the previously deleted `create_widget_setting`. 🙂

lib/chat_api/accounts.ex
adipurnama: Should we delete the `widget_settings, user_profile & user_setting` upon `user / account` deletion ? Yes, I think we should move the logic on db level.

adipurnama:
On the second thought @reichert621 , would it be better if the  `widget_settings, user_profile & user_setting` creation logic moved to the respective `get_<object>_by` method ? E.g. `get_settings_by_account/1` should create new `widget_settings` entry for the account if it doesn't exists yet. That way, we can takeout the `Repo.transaction` block & avoid creating migration for `user` who doesn't have `user_profile` yet.

What do you think?

reichert621:
great idea @adipurnama! i think that would simplify things for now 😄

lib/chat_api/accounts.ex
reichert621: yup, we can handle that in a separate ticket :)

adipurnama:
done, kindly check the updated code @reichert621

lib/chat_api/users.ex
```diff
@@ -109,19 +114,36 @@ defmodule ChatApi.Users do
...
       iex> get_user_profile(123)
       %UserProfile{}
 
-      iex> get_user_profile(456)
-      ** (Ecto.NoResultsError)
-
   """
   def get_user_profile(user_id) do
-    UserProfile |> where(user_id: ^user_id) |> Repo.one() |> Repo.preload(:user)
+    case UserProfile |> where(user_id: ^user_id) |> Repo.one() do
```
reichert621, line 126: nitpick -- this might just be my personal preference, but can we instead just pipe directly into the `case` statement? e.g.
```elixir
UserProfile
|> where(user_id: ^user_id)
|> Repo.one()
|> case do
  %UserProfile{} = profile ->
    profile
    |> Repo.preload(:user)

  nil ->
    create_user_profile(user_id)
end
```
I find it a tiny bit easier to read but might just be me :P

lib/chat_api/users.ex
```diff
@@ -109,19 +114,36 @@ defmodule ChatApi.Users do
...
+        profile
+        |> Repo.preload(:user)
+
+      nil ->
+        create_user_profile(user_id)
+    end
+  end
+
+  defp create_user_profile(user_id) do
+    case %UserProfile{}
```
reichert621, line 137: same as above, e.g.
```elixir
%UserProfile{}
|> UserProfile.changeset(%{user_id: user_id})
|> Repo.insert()
|> case do
  {:ok, profile} ->
    profile
    |> Repo.preload(:user)

  {:error, _reason} ->
    nil
end
```

lib/chat_api/users/user.ex
```diff
@@ -39,7 +39,7 @@ defmodule ChatApi.Users.User do
     |> validate_inclusion(:role, ~w(user admin))
   end
 
-  @spec role_changeset(Ecto.Schema.t() | Ecto.Changeset.t(), map()) :: Ecto.Changeset.t()
+  @spec disabled_at_changeset(Ecto.Schema.t() | Ecto.Changeset.t(), map()) :: Ecto.Changeset.t()
```
reichert621, line 42: oops, thank you!

lib/chat_api/widget_settings.ex
```diff
@@ -38,42 +38,31 @@ defmodule ChatApi.WidgetSettings do
...
-      iex> create_widget_setting(%{field: bad_value})
-      {:error, %Ecto.Changeset{}}
-
-  """
-  def create_widget_setting(attrs \\ %{}) do
-    %WidgetSetting{}
-    |> WidgetSetting.changeset(attrs)
-    |> Repo.insert()
+  defp create_setting_by_account(account_id) do
+    case %WidgetSetting{}
```
reichert621, line 54: (same here re: piping into `case`)

test/support/test_fixture_helpers.ex
```diff
@@ -0,0 +1,162 @@
+defmodule ChatApi.TestFixtureHelpers do
```
reichert621, line 1: woohoo! 🎉 🎉 🎉

reichert621:
@adipurnama just a couple nitpicky comments, but looks great! going to merge and handle the nitpicks myself :)

PR #263, opened 2020-09-30
Add time zone to customers table

### LINKED ISSUE #261: Add `time_zone` field to customers
Seems like it could be useful? (Will need to send it through on the widget side, but for the sake of this ticket just add a nullable string field to the customers table)

See https://stackoverflow.com/questions/1091372/getting-the-clients-timezone-offset-in-javascript

estevanjantsk:
### Description

Add `time_zone` field to customers

### Issue

Resolves #261 

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api/customers/customer.ex
+++ b/lib/chat_api/customers/customer.ex
@@ -29,6 +29,7 @@ defmodule ChatApi.Customers.Customer do
     field(:screen_height, :integer)
     field(:screen_width, :integer)
     field(:lib, :string)
+    field(:time_zone, :string)
 
     # Freeform
     field(:metadata, :map)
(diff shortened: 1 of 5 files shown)


reichert621 (commented):
thanks @estevanjantsk, looks great! 🎉 

there are a couple tiny things that would make this perfect:
- can you add the `time_zone` field to the `customer_view.ex` so that it will show up in the JSON responses?
- can you also add some quick tests to `customers_test.exs` and `customer_controller_test.exs` that basically tests that the `update_customer_metadata` method works correctly when trying to send an update such as `%{time_zone: "America/New_York"}`?

thank you so much! 😄

estevanjantsk:
Sure! Will add that later. Thanks

reichert621:
thanks, I appreciate the help!

estevanjantsk:
@reichert621 🆗 👍  huhuhu

reichert621 (approved):
awesome! 👍

PR #264, opened 2020-09-30
Add count messages per user function

### LINKED ISSUE #257: [Reporting - API] Add method to determine breakdown of messages sent per user on team
**Problem**
Users should be able to see a breakdown of how many messages are sent by users/agents on their team. (This is only relevant for accounts with multiple users/agents on the team.)

**Solution**
Let's create a method in the Reporting context (https://github.com/papercups-io/papercups/blob/master/lib/chat_api/reporting.ex) that counts the number of messages sent per user on the account for a given date range.

Basically this just means counting messages while grouping by `user_id`, and then expanding the `user` object so we can view details like `email`.

The response should look something like this:
```
[
  %{user: %{id: 1, email: "test@test.com"}, count: 28},
  %{user: %{id: 2, email: "example@example.com"}, count: 33},
  # etc
]
```

**Testing**
Please add tests to `test/chat_api/reporting_test.exs` and make sure they pass with `mix test` :) 

**Context**
See https://github.com/papercups-io/papercups/issues/252

estevanjantsk:
### Description

Add method to determine breakdown of messages sent per user on team.
Basically I added `count_messages_per_user` function that returns all messages grouped by the user_id.

### Issue

Resolves #257 

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api/reporting.ex
+++ b/lib/chat_api/reporting.ex
@@ -4,13 +4,22 @@ defmodule ChatApi.Reporting do
   """
 
   import Ecto.Query, warn: false
-  alias ChatApi.{Repo, Conversations.Conversation, Messages.Message}
+  alias ChatApi.{Repo, Conversations.Conversation, Messages.Message, Users.User}
 
   # TODO: filter by records created between a given `from_date` and `to_date`
   def messages_by_date(account_id) do
     Message
     |> where(account_id: ^account_id)
     |> count_grouped_by_date()
+    |> Repo.all()
+  end
+
+  def count_messages_per_user(account_id) do
+    Message
+    |> where(account_id: ^account_id)
+    |> join(:inner, [m], u in User, on: m.user_id == u.id)
+    |> select([m, u], %{user: %{id: u.id, email: u.email}, count: count(m.user_id)})
+    |> group_by([m, u], [m.user_id, u.id])
     |> Repo.all()
   end
 
--- a/test/chat_api/reporting_test.exs
+++ b/test/chat_api/reporting_test.exs
@@ -42,6 +42,38 @@ defmodule ChatApi.ReportingTest do
              ] = Reporting.messages_by_date(account.id)
     end
 
+    test "count_messages_per_user/1 should return correct number of messages sent per user on team",
+         %{
+           account: account,
+           customer: customer
+         } do
+      account_2 = account_fixture()
+      account_3 = account_fixture()
+      user_2 = user_fixture(account_2)
+      user_3 = user_fixture(account_3)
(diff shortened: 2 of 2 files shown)


test/chat_api/reporting_test.exs
```diff
@@ -42,6 +42,38 @@ defmodule ChatApi.ReportingTest do
...
+
+      message_fixture(account, conversation, %{
+        inserted_at: ~N[2020-09-03 12:00:00],
+        user_id: user_3.id
+      })
+
+      assert [
+               %{count: 2},
+               %{count: 1}
+             ] = Reporting.messages_count_by_date(account.id)
```
reichert621, line 74: should this be `Reporting.count_messages_per_user`?

test/chat_api/reporting_test.exs
```diff
@@ -42,6 +42,38 @@ defmodule ChatApi.ReportingTest do
              ] = Reporting.messages_by_date(account.id)
     end
 
+    test "count_messages_per_user/1 should return correct number of messages sent per user on team",
+         %{
+           account: account,
+           customer: customer
+         } do
+      account_2 = account_fixture()
+      account_3 = account_fixture()
```
reichert621, line 51: these extra accounts shouldn't be necessary, we can just add users to the existing `account` from the setup :)

reichert621 (changes requested):
thanks for tackling this as well @estevanjantsk 😁 

the test looks like it won't pass, so just left a couple comments there

lib/chat_api/reporting.ex
```diff
@@ -14,6 +14,15 @@ defmodule ChatApi.Reporting do
     |> Repo.all()
   end
 
+  def count_messages_per_user(account_id) do
```
reichert621, line 17: 🎉 🎉 🎉

test/chat_api/reporting_test.exs
estevanjantsk: Oh, I changed the function name but I forgot to push the latest commit. 😆 I'm sure this works hahaha. At least on my machine haha.

estevanjantsk:
I will push everything after work. Btw this is a cool project 😊.

test/chat_api/reporting_test.exs
reichert621: 🎉

reichert621:
thanks @estevanjantsk! 😊

estevanjantsk:
@reichert621 👌 🆗

reichert621:
looks great! 👍

reichert621 (approved):
thanks for fixing this up!

Issue #256, opened 2020-09-29
[Reporting - API] Add method to calculate average/total received messages by day of the week


reichert621:
**Problem**
Users should be able to see which days of the week (e.g. Monday, Tuesday, etc) are the "busiest" (i.e. receive the most messages on average)

**Solution**
Let's create a method in the Reporting context (https://github.com/papercups-io/papercups/blob/master/lib/chat_api/reporting.ex) that does the following:
- Retrieves daily counts of "received" messages
  - "Received" messages are messages where `customer_id` is not `null` (i.e. it was sent by a customer through the chat widget) 
- Calculates the average number sent by day of the week (e.g. Monday, Tuesday, etc)
- Calculates the total sent by day of week

The response should look something like this:
```
[
  %{day: "Monday", average: 6.6, total: 20},
  %{day: "Tuesday", average: 9.0, total: 27},
  %{day: "Wednesday", average: 10.3, total: 31},
  # etc
]
```

**Testing**
Please add tests to `test/chat_api/reporting_test.exs` and make sure they pass with `mix test` :) 

**Context**
See https://github.com/papercups-io/papercups/issues/252

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

sbacarob:
I'd love to tackle this, if that's ok

reichert621:
certainly, that would be great @sbacarob! assigning you now :)

reichert621:
done in https://github.com/papercups-io/papercups/pull/274

thanks @sbacarob!

PR #274, opened 2020-10-05
[Reporting - API] Add method to calculate average/total received messages by day of the week

sbacarob:
### Description

This PR introduces a function in `reporting.ex` to calculate the average and total of messages per day of week for a given account that have a `customer_id`.

I'm unsure about having a function produce the list of weekdays, I can revisit if necessary.

Added tests for a couple of scenarios.

### Issue

https://github.com/papercups-io/papercups/issues/256

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api/reporting.ex
+++ b/lib/chat_api/reporting.ex
@@ -37,4 +37,30 @@ defmodule ChatApi.Reporting do
     |> select([r], %{date: fragment("date(?)", field(r, ^field)), count: count(r.id)})
     |> order_by([r], asc: fragment("date(?)", field(r, ^field)))
   end
+
+  def messages_by_weekday(account_id) do
+    Message
+    |> where(account_id: ^account_id)
+    |> where([m], not is_nil(m.customer_id))
+    |> count_grouped_by_date()
+    |> select_merge([m], %{day: fragment("to_char(date(?), 'Day')", m.inserted_at)})
+    |> Repo.all()
+    |> Enum.group_by(&String.trim(&1.day))
+    |> compute_weekday_aggregates()
+  end
+
+  defp compute_weekday_aggregates(grouped) do
+    Enum.map(weekdays(), fn weekday ->
+      records = Map.get(grouped, weekday, [])
+      total = Enum.reduce(records, 0, fn x, acc -> x.count + acc end)
+
+      %{
+        day: weekday,
+        average: total / max(length(records), 1),
+        total: total
+      }
+    end)
+  end
+
+  defp weekdays, do: ~w(Monday Tuesday Wednesday Thursday Friday Saturday Sunday)
 end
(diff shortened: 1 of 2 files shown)


lib/chat_api/reporting.ex
```diff
@@ -51,4 +51,30 @@ defmodule ChatApi.Reporting do
...
+
+      %{
+        day: weekday,
+        average: total / max(length(records), 1),
+        total: total
+      }
+    end)
+  end
+
+  defp weekdays, do: ~w(Monday Tuesday Wednesday Thursday Friday Saturday Sunday)
```
reichert621, line 79: agreed that this might belong somewhere else, but i think this is fine for now 👍 (if we decide to move it, it would be trivial anyway)

lib/chat_api/reporting.ex
```diff
@@ -51,4 +51,30 @@ defmodule ChatApi.Reporting do
     |> select([r], %{date: fragment("date(?)", field(r, ^field)), count: count(r.id)})
     |> order_by([r], asc: fragment("date(?)", field(r, ^field)))
   end
+
+  def messages_by_weekday(account_id) do
```
reichert621, line 55: small request -- let's rename this to `count_messages_by_weekday` so that it's clear we're returning the counts and not the actual messages

test/chat_api/reporting_test.exs
```diff
@@ -194,5 +194,132 @@ defmodule ChatApi.ReportingTest do
...
+      assert [
+               %{day: "Monday", average: 0.0, total: 0},
+               %{day: "Tuesday", average: 0.0, total: 0},
+               %{day: "Wednesday", average: 0.0, total: 0},
+               %{day: "Thursday", average: 0.0, total: 0},
+               %{day: "Friday", average: 0.0, total: 0},
+               %{day: "Saturday", average: 0.0, total: 0},
+               %{day: "Sunday", average: 0.0, total: 0}
+             ] = Reporting.messages_by_weekday(account.id)
+    end
```
reichert621, line 323: hooray for all these tests! ❤️ 

if you don't mind, could i ask you a small favor? now that this test file is growing, i think it would be nice to start grouping tests within a `describe` block, e.g.
```elixir
describe "messages_by_weekday/1" do
  test "messages_by_weekday/1 correctly calculates total and avg of customer messages per day" do
    # ...
  end

  # etc etc
end
```

reichert621 (changes requested):
looks great @sbacarob! 🎉 🎉 🎉 

just a couple small requests and then should be good to merge :)

lib/chat_api/reporting.ex
sbacarob: Sure, I'll update it shortly :+1:

test/chat_api/reporting_test.exs
sbacarob: Absolutely! I thought of doing this but was unsure. Thank you for the feedback!

reichert621 (approved):
woohoo thanks for the quick turnaround on this! 😄

PR #277, opened 2020-10-06
Add .gitattributes file

eikooc:
### Description

Add .gitattributes to enforce LF eol as suggested by @lingster


reichert621:
thanks for handling this 👍

Issue #260, opened 2020-09-29
Display customer metadata in customer details modal


reichert621:
**Problem**
Users can set custom metadata on their customers, but we currently don't display this information anywhere in the UI.

This is what we currently display:
<img width="1084" alt="Screen Shot 2020-09-29 at 4 35 11 PM" src="https://user-images.githubusercontent.com/5264279/94613354-79ec7900-0272-11eb-923b-109232295829.png">

**Solution**
Using this `metadata` field on the customer object:
<img width="565" alt="Screen Shot 2020-09-29 at 4 39 35 PM" src="https://user-images.githubusercontent.com/5264279/94613352-79ec7900-0272-11eb-818a-48c947b346ed.png">

Display these fields at the bottom of the customer details modal: https://github.com/papercups-io/papercups/blob/master/assets/src/components/customers/CustomerDetailsModal.tsx#L22

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

lukewaring:
Hi @reichert621, I would be happy to work on this one 🙂

reichert621:
thanks @lukewaring!

sohamsshah:
Hi @reichert621 I am interested to contribute to Papercups. Can you suggest or assign me some work to start contributing towards?

reichert621:
hey @sohamsshah -- this should have some good guidance :) 

https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md

sohamsshah:
@reichert621 Thank you! Looking forward to be a part of this!

eikooc:
Hi. I was just playing around, and thought I'd do this one since there has been a week without activity. @lukewaring I hope I'm not intruding, and if you have made something more elegant, you can tell me and open a PR 🙂.

My PR is here #288

PR #279, opened 2020-10-06
Add customer metadata in customer details modal (#260)

lukewaring:
### Description

Adds the ad hoc customer metadata to the bottom of the customer details modal. More info on how users can add this metadata is available in the September 18, 2020 [entry](https://github.com/papercups-io/papercups/wiki/Changelog#friday-18-september) of the changelog.

### Issue

#260 - Display customer metadata in customer details modal

### Screenshots

Note the highlighted sections, which show that the user can add non-standard date formats. I used the date formats array pattern of Day.js to check whether a valid date is present in each metadata key. More formats can be added to this array over time, if necessary (e.g., accounting for internationalization). I included the most common date formats that came to mind but am happy to adjust.

image

## Checklist

- [X] Everything passes when running `mix test` - 218 tests completed in 34 seconds
- [X] Ran `mix format`

### DIFF
--- a/assets/src/components/customers/CustomerDetailsModal.tsx
+++ b/assets/src/components/customers/CustomerDetailsModal.tsx
@@ -2,10 +2,42 @@ import React from 'react';
 import {Box, Flex} from 'theme-ui';
 import dayjs from 'dayjs';
 import utc from 'dayjs/plugin/utc';
+import {capitalize} from 'lodash';
 import {Button, Modal, Paragraph, Text} from '../common';
 
 // TODO: create date utility methods so we don't have to do this everywhere
 dayjs.extend(utc);
+
+export const CustomerMetadataSection = ({
+  metadata,
+}: {
+  metadata: Record<string, string>;
+}) => {
+  return !metadata ? null : (
+    <React.Fragment>
+      <Box mb={2}>
+        <Box>
+          <Text strong>Custom metadata</Text>
+        </Box>
+      </Box>
+
+      <Flex sx={{justifyContent: 'space-between', flexWrap: 'wrap'}}>
+        {Object.entries(metadata).map(([key, value]: any) => {
+          const label = capitalize(key).split('_').join(' ');
+          return (
+            <Box mb={2} sx={{flex: '0 50%'}} key={key}>
+              <Box>
+                <Text strong>{label}</Text>
+              </Box>
+
+              <Paragraph>{value.toString()}</Paragraph>
+            </Box>
+          );
+        })}
+      </Flex>
+    </React.Fragment>
+  );
+};
 
 export const CustomerDetailsContent = ({customer}: {customer: any}) => {
   const {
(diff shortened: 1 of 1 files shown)


assets/src/components/customers/CustomerDetailsModal.tsx
```diff
@@ -103,6 +104,46 @@ export const CustomerDetailsContent = ({customer}: {customer: any}) => {
...
       </Flex>
+
+      <Box mb={2}>
+        <Box>
+          <Text strong>Metadata</Text>
+        </Box>
+        {Object.entries(metadata).map(([key, value]: any) => {
+          // Note: the array below contains the date formats that will be captured and formatted by dayjs
+          // This approach is based on the examples found here: https://day.js.org/docs/en/parse/string-format
+          const dateFormatsArr = [
```
reichert621, line 115: i think this is an interesting idea, but i'm worried about some confusion here in case this gets parsed incorrectly (e.g. 02/05/2020 could be Feb 5 or May 2), so i'd rather just display whatever string is set in the metadata for now :)

assets/src/components/customers/CustomerDetailsModal.tsx
```diff
@@ -103,6 +104,46 @@ export const CustomerDetailsContent = ({customer}: {customer: any}) => {
...
+            'MM/DD/YY',
+            'DD/MM/YYYY',
+            'DD/MM/YY',
+          ];
+          if (dayjs(value, dateFormatsArr).isValid() === true) {
+            const formattedDate = dayjs.utc(value).format('MMMM DD, YYYY');
+            return (
+              <Paragraph key={key}>
+                {key.charAt(0).toUpperCase() +
+                  key.slice(1).split('_').join(' ')}
```
reichert621, line 132: nice, i like this formatting 👍 

i think we can make it a bit more readable with lodash's `capitalize`, e.g.
```javascript
import {capitalize} from lodash;
// ...
const label = capitalize(key).split('_').join(' ');
```

reichert621 (changes requested):
thanks for tackling this @lukewaring! 😄 

really love the ideas for improving the formatting of the values -- i think for now we should hold of on the date formatting, but the label formatting is great.

in terms of layout, i think i prefer what @eikooc did in https://github.com/papercups-io/papercups/pull/288 -- would you mind updating your PR to mimic that layout, but keep your formatting of the labels?

thank you! ❤️

lukewaring:
Thanks for the feedback @reichert621! I will make these changes later today.

assets/src/components/customers/CustomerDetailsModal.tsx
lukewaring: Ah, good point! There are always so many things that can go wrong with dates/time 😄 I hadn't worked with Day.js before, so it was fun to play around with it a bit.

reichert621:
thanks @lukewaring! excited to merge these updates 🎉

lukewaring:
Hey @reichert621! I just pushed a commit that I believe should get us to where we want to be. But just let me know if I need to make any additional changes. Thank you!

Screen Shot 2020-10-07 at 5 33 30 PM

reichert621 (approved):
great work, thanks @lukewaring! 😁

lukewaring:
Thanks @reichert621, happy to do it!

Issue #286, opened 2020-10-07
Fix auto-assignment logic when integrated with Slack


reichert621:
**Problem**

Since the Slack integration requires all conversations to have an assignee, we auto-assign the first user found on the account whenever a conversation is started over Slack and no assignee has been set.

Now that users can be disabled on an account, it's possible for the auto-assignment to assign a disabled user.

**Solution**
Update this to only assign valid (enabled) users, and default to the earliest created user on the account (based on the `inserted_at` field).

The logic lives here: https://github.com/papercups-io/papercups/blob/master/lib/chat_api/slack.ex#L305-L310

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

reichert621:
@sykrish thanks for handling! going to assign you here as well :)

PR #293, opened 2020-10-07
Fix: Reject disabled users when auto assigning.

sykrish:
### Description

This commit does:
- Rejects users which have a non-nil value assigned to `deleted_at`.
- Sorts the users by `inserted_at`, and gets the oldest users.

### Issue
Resolves: #286 

## Checklist

- [x] Ran `mix format`
- [?] No frontend compilation warnings

### DIFF
--- a/test/chat_api/slack_test.exs
+++ b/test/chat_api/slack_test.exs
@@ -6,7 +6,8 @@ defmodule ChatApi.SlackTest do
   alias ChatApi.{
     Conversations,
     Slack,
-    SlackConversationThreads
+    SlackConversationThreads,
+    Users
   }
 
   describe "slack" do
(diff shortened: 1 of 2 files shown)


test/chat_api/slack_test.exs
```diff
@@ -160,6 +161,23 @@ defmodule ChatApi.SlackTest do
...
+         %{account: account} do
+      {:ok, disabled_user} =
+        account
+        |> user_fixture()
+        |> Users.disable_user()
+
+      primary_user = user_fixture(account)
+
+      # Make sure that secondary_user is inserted later.
+      :timer.sleep(1000)
```
reichert621, line 174: can we do 100ms in stead of a full 1s here? just don't want to slow down tests too much 😛 

```suggestion
      :timer.sleep(100)
```

reichert621 (approved):
thanks @sykrish! 🎉🎉🎉

Issue #148, opened 2020-08-14
Flash number of unread messages on the top of tab 


cheeseblubber:
We currently display the number of unread messages at the top of the tab, but we should enhance this by making it "flash" to draw the user's attention.

Old example:
![118009002_942090216255467_1862770315944140461_n](https://user-images.githubusercontent.com/4218509/90293426-e1707580-de38-11ea-8570-d739519586a1.png)

josephecombs:
(This feature should mimic the way Brex "flashes" the page Title to draw the user's attention)

simran2607:
can u pls elaborate

reichert621:
@simran2607 by "flash" we mean the html title should show `(2) New messages!` (wait 1s) -> `Papercups` (wait 1s) -> `(2) New messages!` -> etc, so that it's more eye catching

Aditya135:
Hi @reichert621 , can you assign this one to me?

reichert621:
sure @Aditya135!

PR #295, opened 2020-10-08
Title flashing

raditya3:
### Description

Added flashing title to notify user about unread conversations

### Issue

[#148 ]

### Screenshots

demo

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/assets/src/components/Dashboard.tsx
+++ b/assets/src/components/Dashboard.tsx
@@ -64,8 +64,7 @@ const Dashboard = (props: RouteComponentProps) => {
   };
 
   useEffect(() => {
-    const interval = setInterval(_toggleNotificationMessage, INTERVAL);
-    return () => clearInterval(interval);
+    const interval = setTimeout(_toggleNotificationMessage, INTERVAL);
   });
 
   return (
--- a/assets/src/components/Dashboard.tsx
+++ b/assets/src/components/Dashboard.tsx
@@ -55,7 +55,7 @@ const Dashboard = (props: RouteComponentProps) => {
 
   const INTERVAL = 2000; //Set flashing interval
 
-  const _toggleNotificationMessage = () => {
+  const toggleNotificationMessage = () => {
     if (notificationMessage.startsWith('New message')) {
       setNotificationMessage('Papercups');
     } else {
@@ -64,7 +64,9 @@ const Dashboard = (props: RouteComponentProps) => {
   };
 
   useEffect(() => {
-    const interval = setTimeout(_toggleNotificationMessage, INTERVAL);
+    if (totalNumUnread > 0) {
+      setTimeout(toggleNotificationMessage, INTERVAL);
+    }
   });
 
   return (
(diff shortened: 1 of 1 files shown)


reichert621:
@Aditya135 can you post a gif of what this update looks like?

also can you verify that when the number of unread messages changes, the html title also changes correctly?

Aditya135:
@reichert621 I have updated the PR comment with the git, can you review it. You might have to zoom it a little bit 😅

assets/src/components/Dashboard.tsx
```diff
@@ -51,14 +51,28 @@ const Dashboard = (props: RouteComponentProps) => {
...
+    if (notificationMessage.startsWith('New message')) {
+      setNotificationMessage('Papercups');
+    } else {
+      setNotificationMessage(`New message${totalNumUnread === 1 ? '' : 's'}`);
+    }
+  };
+
+  useEffect(() => {
+    const interval = setTimeout(_toggleNotificationMessage, INTERVAL);
+  });
```
reichert621, line 68: I'm not too familiar with `useEffect` -- how does this work if there is no second argument here? will every re-render/update trigger this?

also, should we add a `clearTimeout` when this unmounts?

assets/src/components/Dashboard.tsx
```diff
@@ -51,14 +51,28 @@ const Dashboard = (props: RouteComponentProps) => {
 
   const logout = () => auth.logout().then(() => props.history.push('/login'));
 
+  const [notificationMessage, setNotificationMessage] = useState('Papercups');
+
+  const INTERVAL = 2000; //Set flashing interval
+
+  const _toggleNotificationMessage = () => {
```
reichert621, line 58: nitpick -- can we remove the `_` at the beginning of this function name? (it seems unnecessary)

reichert621 (commented):
thanks for including the gif @raditya3 :)

just one last question before we merge!

assets/src/components/Dashboard.tsx
raditya3: ##  useEffect()
`useEffect `will run whenever the component will update/mount.

## clearTimeout()
`clearTimeout` is not necessary as timeout removes itself after INTERVAL is finished so no need to call this function when the component unmounts. [this](https://stackoverflow.com/questions/7391567/when-using-settimeout-do-you-have-to-cleartimeout) for reference. 

## Performance Impact.
It will not impact performance much as only the title tag will be re-rendered. [this](https://stackoverflow.com/questions/24718709/reactjs-does-render-get-called-any-time-setstate-is-called#:~:text=No%2C%20React%20doesn%27t%20render,render%20as%20little%20as%20possible.) for reference.

## Will this be triggered again and again?
yes it will be when the timeout expires and the title changes.

I have tried it with `setInterval` method which resulted in inconsistencies after some time (irregularity in flash interval).

Additional optimization can be that the interval only be created if there are 1 or more unread messages. (will do in next commit)

assets/src/components/Dashboard.tsx
raditya3: I will remove this on next commit

assets/src/components/Dashboard.tsx
reichert621: makes sense, thanks for the notes!

reichert621 (approved):
lgtm! 👍

Issue #287, opened 2020-10-07
[Email notifications] Add customer email to "new message" notification if available


reichert621:
**Problem**
Our email notifications currently don't include any information about the sender. We should include the customer/sender email in the message if it's available.

**Solution**
If the customer sending the message has an email address, display it in the email notification. (Here's where the logic lives: https://github.com/papercups-io/papercups/blob/master/lib/chat_api/messages.ex#L206-L208)

For example, instead of just writing `"A new message has arrived: #{message}"` we should write `"A new message has arrived from #{customer.email}: #{message}"` (in https://github.com/papercups-io/papercups/blob/master/lib/chat_api/emails/email.ex#L19)

**Testing**
Please add tests if possible and make sure everything passes with `mix test` :) 

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

shivamls:
Hi
Please assign this to me.

reichert621:
sure @shivamls :) i'll assign you for now -- when do you think you'll be able to work on it?

shivamls:
Going through the docs and slack. Will start as soon as I get a hang of things (first time contributing)

Targeting completing this by today

reichert621:
sounds good, thanks for the update @shivamls :)

reichert621:
done, thanks @shivamls!

PR #302, opened 2020-10-10
Adding customer info if present to email

shivamls:
### Description

Adding customer email if present in email

### Issue

https://github.com/papercups-io/papercups/issues/287

### Screenshots

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api/emails.ex
+++ b/lib/chat_api/emails.ex
@@ -7,12 +7,13 @@ defmodule ChatApi.Emails do
   alias ChatApi.Emails.Email
   alias ChatApi.Users.{User, UserSettings}
 
-  def send_new_message_alerts(message_obj) do
-    message_obj
+  @spec send_new_message_alerts(Message.t()) :: [any]
+  def send_new_message_alerts(message) do
+    message
     |> Map.get(:account_id)
     |> get_users_to_email()
     |> Enum.map(fn email ->
-      email |> Email.new_message_alert(message_obj) |> deliver()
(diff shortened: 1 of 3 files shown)


lib/chat_api/emails.ex
```diff
@@ -7,11 +7,12 @@ defmodule ChatApi.Emails do
   alias ChatApi.Emails.Email
   alias ChatApi.Users.{User, UserSettings}
 
-  def send_new_message_alerts(message, account_id, conversation_id) do
-    account_id
+  def send_new_message_alerts(message_obj) do
```
reichert621, line 10: small nitpick -- instead of `message_obj` can we just call this `message` and maybe at a typespec above the method to make it clear that this is a Message struct and not just a string, e.g.
```elixir
alias ChatApi.Messages.Message

# ...

@spec send_new_message_alerts(Message.t()) :: [any]
def send_new_message_alerts(message) do
  # ...
end
```

this is something we haven't done a great job at so far, but I'd like to start adding more typespecs moving forward -- it would a great help if you could add some to the methods you updated 😄

reichert621 (commented):
@shivamls looks good! just one small comment :) 

thanks for handling this 🎉

lib/chat_api/emails.ex
shivamls: Sure thing.

reichert621 (approved):
thank you @shivamls! 🎉 🎉 🎉

Issue #297, opened 2020-10-08
Add toggles for `requireEmailUpfront` and `showAgentAvailability` on getting started page on dashboard


cheeseblubber:
**Is your feature request related to a problem? Please describe.**
When a non technical user wants to customize the widget for requiring email upfront and showing agent availability they can't do it through the GUI.

**Describe the solution you'd like**
There should be a toggle for both requireEmailUpFront and showAgentAvailability in the https://github.com/papercups-io/papercups/blob/master/assets/src/components/account/GettingStartedOverview.tsx#L242 section where it generates and the flags for the widget

**Additional context**
This is a customer request from inStock

markholmes:
Should this also incorporate other widget options (as indicated here: https://github.com/papercups-io/chat-widget#options)? And just so I'm certain, this only affects the generated HTML/React installation scripts, correct?

markholmes:
Something like this:

<img width="1680" alt="email_and_availability" src="https://user-images.githubusercontent.com/921826/95669550-f450af00-0b36-11eb-946c-d0e322f19a9b.png">

<img width="675" alt="html" src="https://user-images.githubusercontent.com/921826/95669551-f7e43600-0b36-11eb-9fbe-557abdfdbcc5.png">

<img width="680" alt="react" src="https://user-images.githubusercontent.com/921826/95669554-fa469000-0b36-11eb-8754-c91ace5c57cf.png">

reichert621:
hey @markholmes! awesome, looks pretty good so far :) 

only feedback would be:
- `true` and `false` need to be booleans in the code snippets rather than strings
- you'll also need to apply these updates to the widget on the page so the user can see what it looks like when e.g. `showAgentAvailability` is `true`

let me know if you have any questions! going to assign this to you for now 👍

reichert621:
hey @markholmes! any updates on the widget side of this? (i remember you were working on it, sorry if i missed any latest updates!)

a8t:
Since I'm already looking at this page, I can look at this too.

reichert621:
Oops I believe this is done now! We can close this.

PR #303, opened 2020-10-11
Add email and agent availability options to the Getting Started page

markholmes:
### Problem

When a non technical user wants to customize the widget for requiring email upfront and showing agent availability they can't do it through the GUI.

### Solution

On the "Getting Started" page, add toggles for "Agent Availability" and "Require User Email" and add text inputs for "Agent Available Text" and "Agent Unavailable Text".

### Notes

- Toggling `requireEmailUpfront` doesn't show up within the account preview because the user is already identified.
- Toggling `showAgentAvailability` requires a page refresh — is there any way to prevent this?
- Similarly, when `showAgentAvailability` is set to `true`, changing the color removes the `showAgentAvailability` box on the chat preview, even though the state of the agent availability hasn't changed.

Also, it's worth noting that as we continue to add more customizations to this Getting Started page, it may be worth while to begin to categorize them into separate cards. For example, "Header", "Message Input", "Appearance", etc.

### Issue

https://github.com/papercups-io/papercups/issues/297

### Screenshots

papercups-add-email-and-availability

## Checklist

- [x] Ran `mix format`


reichert621:
hey @markholmes this looks great! thanks for tackling it 😄 

let me look into a couple of the notes you mentioned before merging this, since ideally we would be able to have the changes reflected in the preview right away (but to be clear, this is something I need to fix on our end -- your code is fine!)

markholmes:
@reichert621 If you have any clues on what the issue is and can point me in the right direction, I'd be happy to try to tackle it as well. I just wasn't sure where to look. Is it React/iFrame weirdness?

reichert621:
@markholmes yup that's exactly it -- our "chat widget" (https://github.com/papercups-io/chat-widget) renders our "chat window" (https://github.com/papercups-io/chat-window) in an iframe, and many of these props are passed down via query params.

this might be a good starting point if you wanna try digging into it! https://github.com/papercups-io/chat-widget/blob/master/src/components/ChatWidgetContainer.tsx#L217-L229

(apologies in advance though -- it may be slightly annoying to get all these things running together... i tried to write some docs here https://github.com/papercups-io/chat-widget#development but let me know if you need any helping getting it up and running 😅 )

Issue #281, opened 2020-10-07
[Reporting - API] Add method to determine number of new customers created per day


reichert621:
**Problem**
Users should be able to see how many new customers are interacting with their chat per day.

**Solution**
Let's create a method in the Reporting context (https://github.com/papercups-io/papercups/blob/master/lib/chat_api/reporting.ex) that counts the number of customers created on the account per day for a given date range.

The response should look something like this:
```
[
  %{date: "2020-09-01", count: 7},
  %{date: "2020-09-02", count: 12}, 
  %{date: "2020-09-03", count: 4}
]
```

(This method will be quite similar to the `count_messages_by_date` method: https://github.com/papercups-io/papercups/blob/master/lib/chat_api/reporting.ex#L9)

Also, when this method is ready, feel free to add it to the response payload in the `ReportingController`!

**Testing**
Please add tests to `test/chat_api/reporting_test.exs` and make sure they pass with `mix test` :) 

**Context**
See https://github.com/papercups-io/papercups/issues/252

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

jonathanyeong:
Hey there! Could I take this issue?

reichert621:
sure @jonathanyeong! assigning you now :)

reichert621:
(i should add -- if this ticket is too easy for you, feel free to look out for any opportunities to do some refactoring in the `Reporting` context or in the tests!)

jonathanyeong:
Haha let's see how I go with this ticket! First time working with a prod Elixir app!

m1ome:
@jonathanyeong are you working on this? would like to pick this up if you are not working on it 😄

jonathanyeong:
Hi! @m1ome I've definitely been dragging my feet on this. Looks like you already have a PR open. So go for it!

reichert621:
hey @jonathanyeong feel free to tackle this still! I think the PR misunderstood this a bit because it queries the `users` table rather than the `customers` table

but if you don't have time this week, I'll be writing up some more tickets soon if you'd like to pick up a different one 👍

jonathanyeong:
[...]

PR #312, opened 2020-10-12
Added users by date in reporting api

m1ome:
### Description

Now reporting api will have in response users by date checking #281 

### Issue

https://github.com/papercups-io/papercups/issues/281

## Checklist

- [x] Ran `mix format`


m1ome:
Copy of #314 closing up!

PR #314, opened 2020-10-13
#281 - Add a count for customers created per day

jonathanyeong:
### Description
Adds methods in Reporting to get the number of customers created per day.

Also:
- Add a customer test_changeset. This allows us to modify the `inserted_at` column in tests.
- Add tests for count_customers_by_date methods. Separated from the other test blocks because
  we don't need to create a customer during the setup.

### Issue
https://github.com/papercups-io/papercups/issues/281

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api/customers/customer.ex
+++ b/lib/chat_api/customers/customer.ex
@@ -102,4 +102,10 @@ defmodule ChatApi.Customers.Customer do
       :time_zone
     ])
   end
+
+  def test_changeset(message, attrs) do
+    message
+    |> cast(attrs, [:inserted_at])
+    |> changeset(attrs)
+  end
 end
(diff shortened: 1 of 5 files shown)


lib/chat_api/customers/customer.ex
```diff
@@ -102,4 +102,10 @@ defmodule ChatApi.Customers.Customer do
       :time_zone
     ])
   end
+
+  def test_changeset(message, attrs) do
+    message
```
reichert621, line 107: nit -- let's change `message` to `customer`

reichert621 (commented):
looks great @jonathanyeong! :)  

just one small nitpick and then should be good to merge 🎉

lib/chat_api/customers/customer.ex
jonathanyeong: Argh copy and paste error. Good catch thanks!

reichert621 (approved):
thanks @jonathanyeong! 😄

Issue #282, opened 2020-10-07
[Internal notifications] Send internal Slack alert when user successfully links Slack/Gmail integration


reichert621:
**Problem**
This is a good way for us to track engagement -- if users are setting up these integrations, we want to make sure everything is working well for them. 

**Solution**
In the SlackController and GmailController, send an internal notification to Slack (with `ChatApi.Slack.log`) that says "{user.email} successfully linked Slack/Gmail!"

(See https://github.com/papercups-io/papercups/pull/276 for an example of how this might look)

**NB**: in order to test this manually, you'll need to get our own Slack webhook URL https://api.slack.com/messaging/webhooks and set the `PAPERCUPS_SLACK_WEBHOOK_URL` environment variable to it.

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

lorecrafting:
Hey, thanks for the descriptive and bite-sized issues.  Would you please assign this to me?  Thank you!

reichert621:
sure @lorecrafting!

shivamls:
Hi @lorecrafting if you are not planning to work on this lmk. I can pick this up.

lorecrafting:
Thanks @shivamls, i'll be putting in a PR today.

reichert621:
thanks @lorecrafting!

PR #315, opened 2020-10-13
Notify internal Slack of customer Oauth of Slack and Gmail - Issue #282

lorecrafting:
### Description

Notify's Papercups internal Slack if a customer performs an oauth flow on Slack and Gmail.

### Issue

_https://github.com/papercups-io/papercups/issues/282_

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api_web/controllers/gmail_controller.ex
+++ b/lib/chat_api_web/controllers/gmail_controller.ex
@@ -113,7 +113,7 @@ defmodule ChatApiWeb.GmailController do
         ChatApi.Slack.log("#{email} successfully linked Gmail!")
       end)
     end
+
     conn
   end
-
 end
--- a/lib/chat_api_web/controllers/gmail_controller.ex
+++ b/lib/chat_api_web/controllers/gmail_controller.ex
@@ -75,6 +75,7 @@ defmodule ChatApiWeb.GmailController do
       )
       |> case do
         {:ok, result} ->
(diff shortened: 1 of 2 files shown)


lorecrafting:
I haven't tested it with the #sandbox channel yet, as i'm getting redirect_uri errors.

Would you please let me know if what I currently have is in the right direction? Thanks!

lib/chat_api_web/controllers/gmail_controller.ex
```diff
@@ -75,6 +75,7 @@ defmodule ChatApiWeb.GmailController do
       )
       |> case do
         {:ok, result} ->
+          notify_slack(conn)
```
reichert621, line 78: we can also pipe this with the line below :)
```elixir
conn |> notify_slack() |> json(%{ok: true, data: result})
```

reichert621:
@lorecrafting looks fine so far -- just one suggestion :) 

sorry this will be a little difficult to test, so feel free to just make the update suggested above and then i can handle testing 👍 

also, can you update the PR title to something more descriptive? thanks!

lib/chat_api_web/controllers/gmail_controller.ex
lorecrafting: Done!

reichert621 (approved):
thanks @lorecrafting!

Issue #308, opened 2020-10-11
Fix some `any` types in the Conversations UI


reichert621:
**Is your feature request related to a problem? Please describe.**
We have a lot of `any` types in our frontend. Let's start replacing these with more descriptive types.

**Describe the solution you'd like**
In all the components in `assets/src/components/conversations`, replace as many `any` types with `Conversation`/`Message`/`Account`/`User` types from https://github.com/papercups-io/papercups/blob/master/assets/src/types.ts as possible :) 

(Note that you may need to add/update some of the types in that file.)

**Additional context**
One thing that might be helpful to understand how data looks coming from the API:

If you're trying to figure out what's included in the `Conversation` type, take a look at the `conversation_view.ex` file on the backend to see what the API response payload includes.

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

xapatjb4:
Hey @reichert621 would like to give this a shot. Can you assign this to me?

raditya3:
Hey, Can I get this one?

reichert621:
sure @raditya3! assigning you for now :)

PR #320, opened 2020-10-14
Type conversation

raditya3:
### Description

Added type check in following files

- assets/src/components/conversations/ConversationDetailsSidebar.tsx
- assets/src/components/conversations/ConversationHeader.tsx
- assets/src/components/conversations/ConversationItem.tsx
- assets/src/components/conversations/ConversationMessages.tsx

Modified types in
- assets/src/types.ts:
     Changed Users, Customer, Message, Conversation type

### Issue

#308 

### Screenshots

image

(Note : Console error due to my ad-blocker)

## Checklist

- [x] Ran `mix format`


assets/src/types.ts
```diff
@@ -2,17 +2,32 @@ export type User = {
   id: number;
   email: string;
   display_name?: string;
-  full_name?: string;
+  name?: string;
```
reichert621, line 5: why this change? I'm not sure this is correct 🤔

reichert621 (changes requested):
I'm not sure the change from `full_name` -> `name` is correct, but let me know if I'm missing something!

assets/src/types.ts
raditya3: image

The BE API returns response having Customer object having property as `name` (not `full_name`), so to maintain uniformity (and to replace `any` with `Customer`), I replaced the variable name

reichert621: ah -- `name` is for the `Customer` type, but `full_name` is for the `User` type

raditya3: https://github.com/papercups-io/papercups/blob/4e9c4580faa7641ccc3adb338456cbcfeca0ab17/assets/src/components/conversations/ConversationHeader.tsx#L198
In this line `user.name` was used, that's why I must have changed the property. I think I should keep the property as `full_name` and change the corresponding line

reichert621: ah yes, that line is incorrect 👍

raditya3: Can you review again?

reichert621 (approved):
thanks @raditya3!

Issue #328, opened 2020-10-19
Migrate customers table `current_url`/`pathname` fields from `string` to `text` types


reichert621:
**Problem**
The `current_url` and `pathname` fields on the Customers table are sometimes longer than the max character limit for DB fields of type "string".

**Solution**
Migrate the following fields on the Customers table from type "string" to type "text":
- `current_url`
- `pathname`

For context: https://github.com/papercups-io/papercups/blob/master/lib/chat_api/customers/customer.ex#L29

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

kamilpikula:
if the problem's similar to https://github.com/papercups-io/papercups/issues/329 - you can assign to me. :)

reichert621:
hey @kamilpikula! sounds good :)

kamilpikula:
hi @reichert621, I sent PR with the changes - https://github.com/papercups-io/papercups/pull/338 :)

reichert621:
thanks @kamilpikula!

PR #338, opened 2020-10-19
fix: migrate customers table `current_url`/`pathname` fields from `string` to `text` types

kamilpikula:
### Problem

The current_url and pathname fields on the Customers table are sometimes longer than the max character limit for DB fields of type "string".

### Solution

Migrate the following fields on the Customers table from type "string" to type "text".
Created a migration script and updated this fields at the database level

### Note

`:text` type can be used in migration, but not the schema. In the database `:string` type will be mapped to `text` column type.

### Issue

[#328 ](https://github.com/papercups-io/papercups/issues/328)

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/priv/repo/migrations/20201019210311_change_current_url_and_pathname_fields_type_to_text.exs
+++ b/priv/repo/migrations/20201019210311_change_current_url_and_pathname_fields_type_to_text.exs
@@ -1,0 +1,10 @@
+defmodule ChatApi.Repo.Migrations.ChangeCurrentUrlAndPathnameFieldsTypeToText do
+  use Ecto.Migration
+
+  def change do
+    alter table(:customers) do
+      modify :current_url, :text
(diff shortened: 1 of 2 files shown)


reichert621 (commented):
looks good @kamilpikula! will this affect any of our existing data?

also, do you mind adding some tests to `test/chat_api/customers_test.exs` where we create/update a Customer with a long (i.e. more than 255 char) string for these fields to verify that it works correctly?

thanks! 🎉

kamilpikula:
@reichert621 there shouldn't be any problems as we just changed the field type to `text` from `string`, so you can use more than 255 characters (limit of `string`)

I added the above fields to `update_customer/2` test and this seems to work. :)

reichert621:
awesome, thanks @kamilpikula! 🎉

Issue #329, opened 2020-10-19
Migrate widget_settings table `pathname` fields from `string` to `text` type


reichert621:
**Problem**
The `pathname` field on the `widget_settings` table is sometimes longer than the max character limit for DB fields of type "string".

**Solution**
Migrate the `pathname` fields on the `widget_settings` table from type "string" to type "text":

https://github.com/papercups-io/papercups/blob/master/lib/chat_api/widget_settings/widget_setting.ex#L18

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

yash2189:
hi can I work on this? Also I had a question... do we just need to change the type from `string` to `text`?

reichert621:
hey @yash2189! it will also involve creating a migration script so that this type gets updated at the database level

kamilpikula:
hi, I sent PR with solution: #343 :)

reichert621:
thanks @kamilpikula!

PR #343, opened 2020-10-22
Migrate widget_settings table `pathname` fields from `string` to `text` type

kamilpikula:
### Problem

The pathname field on the widget_settings table is sometimes longer than the max character limit for DB fields of type "string".

### Solution

Migrate the `pathname` fields on the `widget_settings` table from type "string" to type "text.
Created a migration script and updated this fields at the database level.

### Issue

[#329 ](https://github.com/papercups-io/papercups/issues/329)

## Checklist

- [x] Ran `mix format`


reichert621 (approved):
thanks @kamilpikula! 🎉

Issue #334, opened 2020-10-19
Archive conversations that are older than 14 days


cheeseblubber:
**Is your feature request related to a problem? Please describe.**
Currently on the client side we are fetching and listening for changes on all conversations -- even those that were closed month ago. This is starting to slow things down.

Let's add a worker that archives all closed conversations that haven't been updated in over 14 days (2 weeks). This will at least speed things up until we start optimizing our queries (and add pagination, etc).

**Describe the solution you'd like**
We should archive conversations that haven't been updated in 14 days.

* A job should run on a daily basis and archive conversations that haven't been updated in 14 days
  * e.g. `conversation.updated_at` is more than 14 days ago and `conversation.status` is `"closed"`
  * We can use the new `ChatApi.Conversations.archive_conversation` method :) 
* Archived conversations should not show up in the UI anywhere (both in the dashboard and in the chat widget)
  * Edge case would be Slack... maybe we should add a message to the Slack thread that says "this conversation has been archived" or something? (we should probably do that when the conversation is closed as well... I'll set up separate ticket(s) for these issues)

**Additional context**
Archived conversations should not show up anywhere in the UI. For now we should treat them as though they're deleted. The only reason we're not actually deleting them is in case our users need us to restore the data for them.

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

daskycodes:
I could give this a try and prepare a PR for this full-stack feature.

> Life cycle of a conversation: Open -> Archived -> Deleted (we don't have delete yet)
shouldn't it be Open -> Closed -> Archived -> Deleted?

reichert621:
hey @daskycodes! I think I'd like to break this ticket up into smaller tickets before anyone starts working on this... there's a few things we need to do here:
- we should probably add a `closed_at` field so we known precisely when a conversation was closed
  - for all existing closed conversations, we can just default that field to the current time in the migration
- we'll probably want to add an `archived_at` field as well
- I'm not sure how we should be using the `status` field on the conversation... if we have a `closed_at` and `archived_at` field, it might be a bit redundant to have a `status: "closed"` or `status: "archived"` 🤔 
[...]

PR #356, opened 2020-10-28
Add reccuring job to archive conversations that are older than 14 days

daskycodes:
### Description

- Add `query_conversations_closed_for/1`, a function that returns an `Ecto.Query` of conversations that have been closed for the given days
- Add `archive_conversations/1`, a function that archives conversations for the given query
- Add `ChatApi.Workers.ArchiveConversations`, an Oban Worker that runs every hour an archives conversations which have been closed for more than 14 days

### Issue

#334 

## Checklist

- [x] Ran `mix format`


daskycodes:
@reichert621 

I am not 100% Sure about using `|> Enum.each(&Conversations.archive_conversation/1)` in

```elixir
defmodule ChatApi.Workers.ArchiveConversations do
  use Oban.Worker, queue: :default

  alias ChatApi.Conversations

  @impl Oban.Worker
  def perform(%Oban.Job{} = _job) do
    Conversations.list_conversations_to_archive()
    |> Enum.each(&Conversations.archive_conversation/1)
  end
end

```

do you think `Task.async_stream/3` [docs](https://hexdocs.pm/elixir/Task.html#async_stream/3) would be a better option or even something else?

test/chat_api/conversations_test.exs
```diff
@@ -163,5 +163,17 @@ defmodule ChatApi.ConversationsTest do
...
+    test "list_conversations_to_archive/0 returns conversations which have been closed for more than 14 days",
+         %{conversation: conversation, account: account, customer: customer} do
+      date_time = DateTime.add(DateTime.utc_now(), -(14 * 24 * 60 * 60))
+
+      ready_to_archive_conversation =
+        conversation_fixture(account, customer, %{updated_at: date_time, status: "closed"})
+
+      result_ids = Enum.map(Conversations.list_conversations_to_archive(), fn r -> r.id end)
+
+      assert result_ids == [ready_to_archive_conversation.id]
```
reichert621, line 176: should we also test that conversations that have been closed for LESS than 14 days do not show up here?

lib/chat_api/conversations.ex
```diff
@@ -223,6 +223,14 @@ defmodule ChatApi.Conversations do
     archive_conversation(conversation)
   end
 
+  def list_conversations_to_archive() do
+    Conversation
+    |> where([c], is_nil(c.archived_at))
+    |> where(status: "closed")
+    |> where([c], c.updated_at < from_now(14, "day"))
```
reichert621, line 230: should this be `from_now(-14, "day")`? (i've never used `from_now` before 😅 )

lib/workers/archive_conversations.ex
```diff
@@ -0,0 +1,11 @@
+defmodule ChatApi.Workers.ArchiveConversations do
+  use Oban.Worker, queue: :default
+
+  alias ChatApi.Conversations
+
+  @impl Oban.Worker
+  def perform(%Oban.Job{} = _job) do
+    Conversations.list_conversations_to_archive()
+    |> Enum.each(&Conversations.archive_conversation/1)
```
reichert621, line 9: hmm i wonder if we should add a method to bulk-update/archive conversations, i.e. `Conversations.archive_conversations(conversations)`?

should we be using `update_all`, or `Ecto.Multi`, or something like that? i'm honestly not sure... (it might not matter too much in this case?)

reichert621 (changes requested):
thanks for tackling this! just a few questions and comments :)

test/chat_api/conversations_test.exs
daskycodes: Yes, that would definitely help!

lib/chat_api/conversations.ex
daskycodes: You are right. I replaced it with the `ago(14, "day")` query 

There are some really helpful query helpers in the [Ecto.Query.API](https://hexdocs.pm/ecto/Ecto.Query.API.html) Module :)

lib/workers/archive_conversations.ex
daskycodes: thank  you for the hint @reichert621 

I created the `archive_conversations/1` function and feel that is much better
I also added a Logger.info to the Oban Worker that logs the number of archived conversations
The test should have all the cases we need, i think

```elixir
  def archive_conversations(days_closed \\ 14) do
    Conversation
    |> where([c], is_nil(c.archived_at))
    |> where(status: "closed")
    |> where([c], c.updated_at < ago(^days_closed, "day"))
    |> Repo.update_all(set: [archived_at: DateTime.utc_now()])
  end
```

reichert621: nice! :)

lib/chat_api/conversations.ex
```diff
@@ -223,6 +223,14 @@ defmodule ChatApi.Conversations do
     archive_conversation(conversation)
   end
 
+  def archive_conversations(days_closed \\ 14) do
+    Conversation
+    |> where([c], is_nil(c.archived_at))
+    |> where(status: "closed")
+    |> where([c], c.updated_at < ago(^days_closed, "day"))
+    |> Repo.update_all(set: [archived_at: DateTime.utc_now()])
```
reichert621, line 231: nice!

lib/chat_api/conversations.ex
```diff
@@ -223,6 +223,14 @@ defmodule ChatApi.Conversations do
     archive_conversation(conversation)
   end
 
+  def archive_conversations(days_closed \\ 14) do
```
reichert621, line 226: i'm a little worried that it's not obvious what this method is doing just by looking at it -- for example, if i'm reading the code and i see `Conversations.archive_conversations()`, i might think ALL conversations would be archived

i wonder if it would be better to be a bit more explicit about what's going on here... maybe something like
```elixir
def query_conversations_closed_for(days: days) do
  Conversation
  |> where([c], is_nil(c.archived_at))
  |> where(status: "closed")
  |> where([c], c.updated_at < ago(^days, "day"))
end

def archive_conversations(query) do
  Repo.update_all(query, set: [archived_at: DateTime.utc_now()])
end
```
and then we would do something like
```elixir
{n, nil} =
  Conversations.query_conversations_closed_for(days: 14)
  |> Conversations.archive_conversations()
```

what do you think?

lib/workers/archive_conversations.ex
```diff
@@ -0,0 +1,16 @@
...
+
+  require Logger
+
+  alias ChatApi.Conversations
+
+  @impl Oban.Worker
+  def perform(%Oban.Job{} = _job) do
+    {n, nil} = Conversations.archive_conversations()
+
+    Logger.info("Archived #{n} conversations")
```
reichert621, line 12: 👍

lib/chat_api/conversations.ex
reichert621: we could also have a method called something like `Conversations.archive_conversations_closed_for(days: days)`, i just want to make sure it's explicit which conversations are being archived :)

reichert621 (changes requested):
hey @daskycodes looks great 😄 just one suggestion!

lib/chat_api/conversations.ex
daskycodes: @reichert621 :D
I like it, not only for being more explicit but also for being able to compose it more loosely with other querys for example :+1:

reichert621 (approved):
awesome, thanks @daskycodes!!! 🎉 ❤️

Issue #304, opened 2020-10-11
[Reporting - API] Add method to determine breakdown of customer browsers/timezones/OS's


reichert621:
**Problem**
Users should be able to see a breakdown of their customers' browser/OS/timezone in the Reporting Dashboard

**Solution**
Let's create a method in the Reporting context (https://github.com/papercups-io/papercups/blob/master/lib/chat_api/reporting.ex) that counts the number of customer browsers/OS's/timezones for a given date range.

Basically this just means grouping the `customers` table by a given field (e.g. `:browser`/`:os`/`:time_zone`) and counting the number of occurrences.

The method might look something like:
```elixir
def get_customer_breakdown(account_id, field, filters \\ %{}) do
  # `field` will be `:os`, `:browser`, `:time_zone`, etc (fields on the Customer model)
  # `filters` should include a `from_date` and `to_date`
end 
```

The response should look something like this:
```elixir
get_customer_breakdown(account_id, :browser, %{from_date: ..., to_date: ...})
# => 
[
  %{browser: "Chrome", count: 25},
  %{browser: "Firefox", count: 12},
  # etc
]
```

**Testing**
Please add tests to `test/chat_api/reporting_test.exs` and make sure they pass with `mix test` :) 

**Context**
See https://github.com/papercups-io/papercups/issues/252

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

shivamls:
You can assign this to me.

reichert621:
hey @shivamls -- any updates?

eikooc:
Does it make sense for me to pick this up? It looks stale.

reichert621:
yup go for it @eikooc :)

eikooc:
@reichert621 I have created a draft pull request at #361. I couldn't for the life of me figure out how to make the field an atom or what is needed to make the key of the map dynamic based on the `field` given as input, so I chose to go with a generic name called `field` for now. If you can help me, then it would be appreciated :)

reichert621:
[...]

PR #361, opened 2020-11-01
Feature: breakdown by fields on Customer model

eikooc:
### Description

Added a function to get the count of customers broken down by fields.

### Issue

#304

## Checklist

- [X] Ran `mix format`

### DIFF
--- a/lib/chat_api/reporting.ex
+++ b/lib/chat_api/reporting.ex
@@ -127,10 +127,10 @@ defmodule ChatApi.Reporting do
     |> where(account_id: ^account_id)
     |> where(^filter_where(filters))
     |> group_by([r], field(r, ^field))
-    # |> select([r], {field(r, ^field), count(r.id)})
-    |> select([r], %{:field => field(r, ^field), :count => count(r.id)})
+    |> select([r], {field(r, ^field), count(r.id)})
     |> order_by([r], desc: count(r.id))
     |> Repo.all()
+    |>  Enum.map(fn {value, count} -> %{field => value, :count => count} end)
   end
 
   # Pulled from https://hexdocs.pm/ecto/dynamic-queries.html#building-dynamic-queries
--- a/lib/chat_api/reporting.ex
+++ b/lib/chat_api/reporting.ex
@@ -120,6 +120,31 @@ defmodule ChatApi.Reporting do
     end)
   end
 
+  # get_customer_breakdown(account_id, :browser, %{from_date: ..., to_date: ...})
+  ## =>
+  # [
+  #  %{browser: "Chrome", count: 25},
+  #  %{browser: "Firefox", count: 12},
+  #  # etc
+  # ]
+  # @spec get_customer_breakdown(binary(), map()) :: [aggregate_by_date()]
+  def get_customer_breakdown(account_id, field, filters \\ %{}) do
+    Customer
+    |> where(account_id: ^account_id)
+    |> where(^filter_where(filters))
+    |> group_by([r], field(r, ^field))
+    # |> select([r], {field(r, ^field), count(r.id)})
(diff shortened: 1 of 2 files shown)


lib/chat_api/reporting.ex
```diff
@@ -120,6 +121,18 @@ defmodule ChatApi.Reporting do
...
+  def get_customer_breakdown(account_id, field, filters \\ %{}) do
+    Customer
+    |> where(account_id: ^account_id)
+    |> where(^filter_where(filters))
+    |> group_by([r], field(r, ^field))
+    # |> select([r], {field(r, ^field), count(r.id)})
+    |> select([r], %{:field => field(r, ^field), :count => count(r.id)})
+    |> order_by([r], desc: count(r.id))
+    |> Repo.all()
+  end
```
reichert621, line 134: @eikooc let's just do something like this for now :) 

```elixir
  @spec get_customer_breakdown(binary(), atom(), map()) :: [aggregate_by_field()]
  def get_customer_breakdown(account_id, field, filters \\ %{}) do
    Customer
    |> where(account_id: ^account_id)
    |> where(^filter_where(filters))
    |> group_by([r], field(r, ^field))
    |> select([r], {field(r, ^field), count(r.id)})
    |> order_by([r], desc: count(r.id))
    |> Repo.all()
    |> Enum.map(fn {value, count} -> %{field => value, :count => count} end)
  end
```

what do you think?

eikooc: That could definitely work. I'll look at it and update.

eikooc: We could also do `Enum.into(%{})` and get `%{"Chrome" => 1, "Firefox" => 1, "Safari" => 1}` as output. What do you think about that?

reichert621: hmm 🤔 I think I prefer the named properties for now!

lib/chat_api/reporting.ex
```diff
@@ -120,6 +121,18 @@ defmodule ChatApi.Reporting do
     end)
   end
 
+  @spec get_customer_breakdown(binary(), binary(), map()) :: [aggregate_by_field()]
```
reichert621, line 124: nit - i think the 2nd arg should be `atom()` right?

lib/chat_api/reporting.ex
```diff
@@ -120,6 +121,18 @@ defmodule ChatApi.Reporting do
...
+  def get_customer_breakdown(account_id, field, filters \\ %{}) do
+    Customer
+    |> where(account_id: ^account_id)
+    |> where(^filter_where(filters))
+    |> group_by([r], field(r, ^field))
+    |> select([r], {field(r, ^field), count(r.id)})
+    |> order_by([r], desc: count(r.id))
+    |> Repo.all()
+    |>  Enum.map(fn {value, count} -> %{field => value, :count => count} end)
```
reichert621, line 133: nit - looks like there's an extra space in front of `Enum`

reichert621 (changes requested):
just a couple small nits and then should be good to go! 😄

lib/chat_api/reporting.ex
reichert621, line 133:
```suggestion
    |> Enum.map(fn {value, count} -> %{field => value, :count => count} end)
```

lib/chat_api/reporting.ex
reichert621, line 124:
```suggestion
  @spec get_customer_breakdown(binary(), atom(), map()) :: [aggregate_by_field()]
```

eikooc:
Sorry for not doing this. I hope it was easy to do the changes by yourself :raised_hands:

reichert621:
@eikooc yup it was just small things, no prob! 😄 

thanks for helping with this!

Issue #362, opened 2020-11-01
Migrations fail on initial setup


jalford14:
**Describe the bug**
When running `mix ecto.setup`, the migration fails on `20200831011352_set_messages_seen_at.exs` producing the error `(Postgrex.Error) ERROR 42703 (undefined_column) column c0.archived_at does not exist`.

**To Reproduce**
1. `mix ecto.reset` (if you dare)
2. See error

**Expected behavior**
All of the migration files to run.

**Screenshots**
I'll save you the trouble of having to drop your DB:
![image](https://user-images.githubusercontent.com/7582183/97804961-08ad3680-1c21-11eb-9503-6c8d61b63ec8.png)

**Additional comments**
I believe that the issue is tied to `lib/chat_api/conversations/conversation.ex:20` where the field `archived_at` is being declared. Since the column isn't added until a later migration, Ecto throws an error. Seeing as this file seems to just be a backfill and no longer helpful to future contributors, I just commented out everything in `def up` as a quick workaround but I don't know if that would be an acceptable solution. I'm really interested in what you guys are building and I'm happy to help in any way!

reichert621:
thanks for flagging this @jalford14!

should be fixed in https://github.com/papercups-io/papercups/pull/363 :)

jalford14:
No problem @reichert621. However, you might still run into issues in the future if you change the shape of `Message`. I don't believe preloading messages will work if you have a field in your message schema that isn't yet present in the database (line 13).

reichert621:
good point! I'll create a separate ticket for that -- just wanted to do the minimal amount of work to unblock current migrations in the meantime :P

PR #363, opened 2020-11-01
Fix `SetMessagesSeenAt` migration

reichert621:
### Description

Fixes the `SetMessagesSeenAt` migration

### Issue

Fixes https://github.com/papercups-io/papercups/issues/362

## Checklist

- [x] Ran `mix format`


priv/repo/migrations/20200831011352_set_messages_seen_at.exs
```diff
@@ -9,6 +9,7 @@ defmodule ChatApi.Repo.Migrations.SetMessagesSeenAt do
 
   def up do
     Conversation
+    |> select([:id])
```
reichert621, line 12: (by default this does a `select(*)`, which was causing issues)

Issue #354, opened 2020-10-28
Send conversation state updates to Slack conversation thread


reichert621:
**Problem**
We don't send any notification to our Slack integration when a conversation is closed, prioritized, archived, etc.

This makes it hard to know the state of a conversation if you're only looking at it in Slack.

**Solution**
Let's send the following messages to the Slack conversation thread when necessary:
- Conversation is closed: _This conversation has been closed._
- Conversation is reopened: _This conversation has been reopened._
- Conversation is archived: _This conversation has been archived._
- Conversation is deleted: _This conversation has been deleted._
- Conversation is marked a priority: _This conversation has been prioritized._
- Conversation is marked a non-priority: _This conversation has been de-prioritized._

Note that the messages should be formatted in italics (or some other different styling) in the markdown sent to Slack in order to make it clear that it ISN'T a message from a user.

**Additional context**
Eventually we'll want to make it possible to close/reopen/prioritize/etc a conversation from Slack, but that's not within the scope of this ticket.

**Testing**
You'll need to set up our Slack integration in order to test this. But please also add some unit/integration tests as well if you can!

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

jalford14:
Mind if I take a stab at this?

reichert621:
Sure @jalford14! It might be slightly annoying because setting up Slack locally is kind of a pain, so let me know if you need any help :) 

Here are the docs for Slack btw: https://github.com/papercups-io/papercups/wiki/Setting-up-Slack

PR #368, opened 2020-11-03
Send conversation state updates to Slack

jalford14:
### Description

Added the ability to send any notification to our Slack integration when a conversation is closed, prioritized, archived or deleted.

### Issue

[354](https://github.com/papercups-io/papercups/issues/354)

### Screenshots

image

## Checklist

- [x] Ran `mix format`


test/chat_api/widget_settings_test.exs
```diff
@@ -6,12 +6,6 @@ defmodule ChatApi.WidgetSettingsTest do
   describe "widget_settings" do
     alias ChatApi.WidgetSettings.WidgetSetting
 
-    @valid_attrs %{
-      color: "some color",
-      subtitle: "some subtitle",
-      title: "some title"
-    }
-
```
jalford14, line 14: Removes a warning when running tests. This is unused.

reichert621 (commented):
thanks for tackling this @jalford14! I'm a bit busy this afternoon, so I'll review this evening :) 

(at a glance, I think one thing that would be nice is decoupling the notification methods from the `Conversation` update/archive methods)

lib/chat_api/conversations/helpers.ex
```diff
@@ -0,0 +1,54 @@
...
+        type: :convo_update
+      )
+
+      {:ok, conversation_state_message}
+    else
+      {:error, "state_invalid"}
+    end
+  end
+
+  def send_multiple_archived_updates(conversations) do
```
reichert621, line 22: is this used anywhere? if not, let's just remove it

lib/chat_api/slack.ex
```diff
@@ -247,6 +247,7 @@ defmodule ChatApi.Slack do
     case type do
       :agent -> "*:female-technologist: Agent*: #{text}"
       :customer -> "*:wave: #{identify_customer(customer)}*: #{text}"
+      :convo_update -> "_#{text}_"
```
reichert621, line 250: nit - let's just spell this out: `:conversation_update`

lib/chat_api/conversations/helpers.ex
```diff
@@ -0,0 +1,54 @@
+defmodule ChatApi.Conversations.Helpers do
+  @moduledoc false
+
+  alias ChatApi.Slack
+
+  def send_conversation_state_update(conversation, state) do
```
reichert621, line 6: let's add typespecs to all these methods :)

lib/chat_api/conversations.ex
```diff
@@ -152,6 +152,8 @@ defmodule ChatApi.Conversations do
 
   """
   def update_conversation(%Conversation{} = conversation, attrs) do
+    Helpers.send_conversation_state_update(conversation, attrs)
```
reichert621, line 155: let's call this separately after the `update_conversation` succeeds, rather than embedding it here within the method itself

(same for the methods below)

reichert621 (changes requested):
thanks for tackling this @jalford14! just a few comments :)

lib/chat_api/conversations/helpers.ex
jalford14: that's supposed to be in use. looks like I missed that 😬

reichert621:
hey @jalford14 any updates on this? :) let me know if you need help with anything!

jalford14:
hey @reichert621! busy couple of days of work but I'm starting back up right now. i'll let you know if I'm running into any issues

lib/chat_api/conversations.ex
```diff
@@ -214,6 +214,8 @@ defmodule ChatApi.Conversations do
   @spec archive_conversation(Conversation.t() | binary()) ::
           {:error, Ecto.Changeset.t()} | {:ok, Conversation.t()}
   def archive_conversation(%Conversation{} = conversation) do
+    Helpers.send_conversation_state_update(conversation, %{"status" => "archived"})
+
     update_conversation(conversation, %{archived_at: DateTime.utc_now()})
```
jalford14, line 219: Is `archive_conversation` used anywhere in the app besides testing? I see that we use `archive_conversations` in an Oban worker but not the singular version of the method. If it's just for testing purposes, I say we keep it here in order to manually make the slack message appear. This was the only way I was able to reproduce the screenshot above.

lib/chat_api/conversations/helpers.ex
```diff
@@ -0,0 +1,59 @@
...
+  @spec send_multiple_archived_updates([Conversation.t()]) :: list()
+  def send_multiple_archived_updates(conversations) do
+    state = %{"status" => "archived"}
+
+    Enum.map(conversations, fn conversation ->
+      send_conversation_state_update(conversation, state)
+    end)
+  end
+
+  defp get_conversation_state_message(state) do
```
jalford14, line 35: Purposely did not add a type spec to the private method. Don't know what your preference is, but I can change that if need be.

jalford14:
I added some emojis to our messages for visual clarity. Totally open to any changes you want to make from a design perspective but I tried to keep it similar to the icons used in-app. Let me know what you think!

reichert621:
love the emojis @jalford14 😄 will take a look at this today!

lib/chat_api/conversations.ex
reichert621: hmm yeah i don't think this method is used anywhere anymore... happy to keep this here for testing in the meantime

lib/chat_api/conversations/helpers.ex
reichert621: i'd appreciate a typespec if you don't mind! could just do `map() -> binary()`, no need to enumerate all the valid states :)

lib/chat_api/conversations/helpers.ex
```diff
@@ -0,0 +1,59 @@
...
+    case state do
+      %{"status" => "open"} ->
+        ":outbox_tray: This conversation has been reopened."
+
+      %{"status" => "closed"} ->
+        ":white_check_mark: This conversation has been closed."
+
+      %{"status" => "archived"} ->
```
reichert621, line 43: hmm i wonder if we should use a different term here since "archived" and "deleted" aren't valid statuses (like "open" and "closed")... maybe `"state" => "archived"` instead? just trying to avoid confusion :P

lib/chat_api_web/controllers/conversation_controller.ex
```diff
@@ -147,7 +148,9 @@ defmodule ChatApiWeb.ConversationController do
   def delete(conn, %{"id" => id}) do
     conversation = Conversations.get_conversation!(id)
 
-    with {:ok, %Conversation{}} <- Conversations.delete_conversation(conversation) do
+    with {:ok, _} <-
+           Helpers.send_conversation_state_update(conversation, %{"status" => "deleted"}),
```
reichert621, line 152: does this have to come first because it won't work after the conversation is deleted?

if yes, mind adding a comment in the code explaining that? it might not be intuitive to future readers of the code :) (i would normally expect the state update notification to come _after_ the operation is successfully performed)

lib/chat_api_web/controllers/conversation_controller.ex
```diff
@@ -139,6 +139,7 @@ defmodule ChatApiWeb.ConversationController do
 
     with {:ok, %Conversation{} = conversation} <-
            Conversations.update_conversation(conversation, conversation_params) do
+      Helpers.send_conversation_state_update(conversation, conversation_params)
```
reichert621, line 142: can we wrap these calls in a `Task.start` so that they run async? i think i'd rather not block on these for now, and i also think it's not too big a deal if the call fails... what do you think?

reichert621 (commented):
thanks for the updates @jalford14 🎉  just a couple tiny comments, but looking great!

lib/chat_api/conversations/helpers.ex
jalford14: ah that's a good point!

lib/chat_api_web/controllers/conversation_controller.ex
jalford14: yeah, that's exactly the reason why. there's no conversation to send to after it gets deleted. i'll add a comment!

lib/chat_api_web/controllers/conversation_controller.ex
jalford14: yeah a slack notification doesn't seem mission-critical so I'm fine if it fails as well.

lib/chat_api/slack.ex
```diff
@@ -197,7 +197,7 @@ defmodule ChatApi.Slack do
     case SlackAuthorizations.get_authorization_by_account(account_id) do
       # Supports a fallback access token as an env variable to make it easier to
       # test locally (assumes the existence of a "bots" channel in your workspace)
-      nil -> %{access_token: get_default_access_token(), channel: "#bots"}
+      nil -> %{access_token: get_default_access_token(), channel: "#bots", channel_id: "1"}
```
jalford14, line 200: Probably need to add tests around this. I had a couple of my tests fail here since `send_conversation_message_alert` is expecting a map that includes a `channel_id` (line 136).

reichert621: makes sense, good catch!

reichert621 (approved):
thanks for all this @jalford14! 🎉 

I'm gonna do a bit of testing in staging before deploying, but looks good :) excited for these updates

jalford14:
@reichert621 sounds good!

reichert621:
thanks again @jalford14, works great! 😁

Issue #181, opened 2020-08-22
Supporting uploading images/attachments to messages in chat


reichert621:
A lot of people have requested the ability to send screenshots in particular. I guess a workaround we can recommend in the meantime is to just upload images to certain websites and share the links, but we should definitely support uploading images/attachments to messages at some point.

alexanderadam:
What is the difference to #130? :thinking: 

**PS:** Thank you so much for papercups! :pray:

reichert621:
good point, I'll remove the duplicate 👍

fkotsian:
Phoenix now supports file upload natively in LiveView!: https://www.phoenixframework.org/blog/phoenix-live-view-upload-deep-dive

PR #373, opened 2020-11-04
File upload for dashboard

cheeseblubber:
### Description

Api for file uploads

### Issue

https://github.com/papercups-io/papercups/issues/181

### Screenshots

## Checklist

- [x] Ran `mix format`


lib/chat_api_web/controllers/registration_controller.ex
```diff
@@ -26,7 +26,7 @@ defmodule ChatApiWeb.RegistrationController do
             # # obvious that a user invitation expires after one use.
             # ChatApi.UserInvitations.expire_user_invitation(invite)
             conn
-            |> send_registration_event()
+            |> send_registration_event(params["company_name"])
```
reichert621, line 29: looks like these changes are already in master, can you rebase to clean up the diff?

lib/chat_api_web/controllers/upload_controller.ex
```diff
@@ -0,0 +1,68 @@
...
+  use ChatApiWeb, :controller
+
+  alias ChatApi.Uploads
+  alias ChatApi.Uploads.Upload
+
+  action_fallback ChatApiWeb.FallbackController
+
+  def create(conn, %{
+        "upload" => upload_params,
+        "message_id" => message_id
```
reichert621, line 11: i'm a little confused about how this API should be used -- does a message need to be created before a file can be uploaded? (it seems like the "uploads" table requires a message_id?)

i wonder if a better model would be something like:
- a `files` table that basically just points at a resource in S3 and has some metadata
- an `attachments` table which handles mapping a `file` to a `message`

then, attaching a file to a message has two steps:
 1. upload a file to create a `file` record
 2. when a message is "sent", this causes the file to be "attached" to the message (in the form of creating an `attachment` record)

let me know if i'm overthinking it though

reichert621 (changes requested):
looks good, mainly a little confused about how the API should be used

lib/chat_api/uploads.ex
```diff
@@ -0,0 +1,60 @@
+defmodule ChatApi.Uploads do
+  @moduledoc """
+  The Uploads context.
+  """
+
+  import Ecto.Query, warn: false
```
WLSF, line 6: I'm not sure we're using Queries in this file yet, since we are only dealing with Repo and Schema changeset, guess we can avoid importing it here for now...

priv/repo/migrations/20201103193047_create_upload.exs
```diff
@@ -0,0 +1,17 @@
+defmodule ChatApi.Repo.Migrations.CreateUpload do
+  use Ecto.Migration
+
+  def change do
+    create table(:upload, primary_key: false) do
```
WLSF, line 5: What about keeping the bounded-context name as `Uploads` and the Schema itself as `File`?

lib/chat_api_web/controllers/upload_controller.ex
```diff
@@ -0,0 +1,68 @@
...
+    {:ok, image_binary} = File.read(upload_params.path)
+
+    bucket_name = System.get_env("BUCKET_NAME")
+
+    if bucket_name == nil do
+      raise "s3 bucket is not specified"
+    end
+
+    upload_result =
+      ExAws.S3.put_object(bucket_name, unique_filename, image_binary)
```
WLSF, line 26: What if we manage to create an isolated module for AWS insertions/deletions/etc...?

Then we would be able to avoid System calls, UUID and File readings through the controller action, it could be like a public API for controllers, like:

```
def create(conn, params) do
  with {:ok, file} <- S3.insert_file(params),
        {:ok, file} <- Uploads.create_file(params) do
    conn
    |> put_status(:created)
    |> render(...)
  end
end
```
(It's just an illustration, you may already have a better idea of how to accomplish this.)

WLSF (commented):
Hey it seems really good! 🎉 🚀 

I've just added a couple of suggestions that may help us improve the code readability.

fkotsian:
FYI Phoenix just added a native API for file upload on 11/20, dunno if it's useful? https://www.phoenixframework.org/blog/phoenix-live-view-upload-deep-dive

reichert621:
@fkotsian we're not using LiveView, so unfortunately it's not relevant to us 😞

lib/chat_api_web/controllers/upload_controller.ex
cheeseblubber: I like the having the attachment being separate step since it is a clearer and it makes the API more explicit. But lets not use the word files since both Elixir and AWS has a concept of a file and it might be conflated. Happy to rename it to something else though.

cheeseblubber:
c46af0ea40204ce8a8a8478cb8c6cc0d

assets/src/components/conversations/ChatMessage.tsx
```diff
@@ -123,6 +124,8 @@ const ChatMessage = ({
   const customerId = customer && customer.id;
   const color = getColorByUuid(customerId);
 
+  // TODO: might be nice to push the boolean logic related to color down to the ChatMessageBox
```
reichert621, line 127: :+1: agreed

Issue #250, opened 2020-09-27
As a agent I want the ability Selecting multiple conversations and marking them as read


cheeseblubber:
**Is your feature request related to a problem? Please describe.**
Closing a large number of conversations is slow

**Describe the solution you'd like**
have checkboxes next to on the left conversations and when you highlight multiple conversations you can have a button that closes them all. like how in gmail you can select multiple and click close on all.

ZmagoD:
Hi, I would like to take over this one.

reichert621:
sounds good @ZmagoD! we still need to sort out the UX a bit, but I'll assign you for now 😄

PR #382, opened 2020-11-09
Mark multiple conversations as closed (WIP)

ZmagoD:
### Description

- Added checkboxes to every conversation, 
- Store the conversation id in an array

### Issue

https://github.com/papercups-io/papercups/issues/250

### Screenshots

WIP

Screenshot 2020-11-09 at 22 17 11

## Checklist

- [ ] Ran `mix format`

### DIFF
--- a/assets/src/components/conversations/ConversationItem.tsx
+++ b/assets/src/components/conversations/ConversationItem.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import {Box, Flex} from 'theme-ui';
 import dayjs from 'dayjs';
 import utc from 'dayjs/plugin/utc';
-import {colors, Badge, Text} from '../common';
+import {Badge, Checkbox, colors, Text} from '../common';
 import {SmileTwoTone, StarFilled} from '../icons';
 import {formatRelativeTime} from '../../utils';
 import {Conversation, Message} from '../../types';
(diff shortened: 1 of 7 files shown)


reichert621:
thanks for getting started on this @ZmagoD! I think we'll definitely need to think about how to make this look nicer from a UI/UX perspective... might be worth taking a look at some other similar products (like Intercom) to see how they handle this

ZmagoD:
HI @reichert621 when working on this I found a TODO comment and I'm a bit confused how to approach this task now
Screenshot 2020-12-01 at 21 36 59

Should we fire multiple queries or do a bulk update of selected items 🤔 

My idea was to create an endpoint that will handle status changes for multiple conversations but I'm not sure if it is a good idea anymore 🤔

reichert621:
hey @ZmagoD sorry i missed this!

i would need to think about this a little bit more (it's currently a low priority 😓) -- but i think ideally we would do a "bulk update" to close them (i think this is how you were thinking?)

in the "bulk closing" case, we'll probably want to change the notification that pops up though... we definitely don't want to show a little notification toast for every single conversation that was closed!

ZmagoD:
Hey @reichert621 I'm really sorry but I can't continue to work on this at the moment so I will close this PR if someone else wants to pick up this task.

PR #392, opened 2020-11-11
Update config.ts

synscen:
Hi, I'd like to bring to your attention, maybe the port should be '3000'.
Thanks.

### Description

_Please describe the changes you made._

### Issue

_Link to the issue (if available)_

### Screenshots

## Checklist

- [ ] Ran `mix format`

### DIFF
--- a/assets/src/config.ts
+++ b/assets/src/config.ts
@@ -13,7 +13,7 @@ export const isHostedProd = window.location.hostname === 'app.papercups.io';
 export const REACT_URL = process.env.REACT_APP_URL || 'app.papercups.io';
 
 export const BASE_URL = isDev
-  ? 'http://localhost:4000'
+  ? 'http://localhost:3000'
   : `https://${REACT_URL}`;
 
 // Defaults to Papercups client ID (it's ok for this value to be public)


reichert621:
thanks @synscen! i think there might be a case where we actually need port 4000, so let me double check before merging this :)

reichert621:
hey @synscen -- looks like we need the port to be `4000`, since that's where the API runs in dev (a few things break when I switch to `3000`)

however, if you're noticing something breaking with `4000`, let me know -- seems like there are probably some improvements we can make here

synscen:
Hi, @reichert621 , thank you for your comment again. I should make it clear about what I find to avoid a confusing comment that takes your time so much. sorry.

Yes, you are right, I just find that the port of baseUrl: "http://localhost:4000" and src="http://localhost:4000/widget.js" in  '**Usage in HTML**' , 'Getting Started page' maybe should be '3000'. 

I wish I've made things clear this time. thanks.

reichert621:
no worries @synscen!

going to close this since the issue here is a bit more nuanced and merging this would break some things, but feel free to open an issue (rather than a PR) if you'd like to discuss :)

Issue #389, opened 2020-11-11
Removed unused variables in tests


reichert621:
**Problem**
When running `mix test`, I see a lot of warnings for unused variables.

**Solution**
Let's either remove these variables, or use them if necessary.

**Testing**
Make sure `mix test` still passes!

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

edgrknye:
Hey @reichert621, I would love to take this up.

reichert621:
go for it @edgrknye! since this is a pretty small issue i'll take the first PR from whoever gets this done :)

PR #393, opened 2020-11-11
Silence warnings in tests

edgrknye:
### Description
Most test assertions pattern matched on variables. These variables, for some assertions, may not necessarily be referenced. 
Because they are referenced, this raises warnings about the variables being ununsed. The raised warnings are in this form
```
warning: variable "renewed_access_token" is unused (if the variable is not meant to be used, prefix it with an underscore)
  test/chat_api_web/api_auth_plug_test.exs:36: ChatApiWeb.APIAuthPlugTest."test can create, fetch, renew, and delete session"/1
```

This PR silences these warnings.

### Issue

#389 

### Screenshots

N/A

## Checklist

- [X] Ran `mix format`

### DIFF
--- a/test/chat_api/slack_test.exs
+++ b/test/chat_api/slack_test.exs
@@ -91,7 +91,7 @@ defmodule ChatApi.SlackTest do
       text = "Hello world"
 
       assert %{
-               "blocks" => blocks,
+               "blocks" => _blocks,
                "channel" => ^channel
              } =
                Slack.get_message_payload(text, %{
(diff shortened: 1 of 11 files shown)


edgrknye:
The test which is failing, I believe, is related to what is discussed [here](https://elixirforum.com/t/strange-dbconnection-connectionerror-with-ecto-3-0-6/19806/6)

reichert621:
thanks @edgrknye! don't worry about that test for now, that's something we need to fix on our end 👍

edgrknye:
Alright. Thank you for clarifying this.

Issue #316, opened 2020-10-13
Move to ex_machina for fixture testing


m1ome:
**Is your feature request related to a problem? Please describe.**
Currently papercups using stuff like self-bound factories to run tests, bad thing about it: 
- It generates some weird functions like: `create_test_user` e.t.c
- It should also be handled in future development

**Describe the solution you'd like**
Solution is to use [ex_machina](https://github.com/thoughtbot/ex_machina) and remove all fixtures generation/generation code out of repository.

**Describe alternatives you've considered**
https://github.com/batate/blacksmith and a couple of others, seems very inmature and hard to config/maintain

reichert621:
great idea @m1ome! this sounds related to https://github.com/papercups-io/papercups/issues/60, do you mind commenting on that issue as well in case these should be done at the same time?

m1ome:
@reichert621 can you assign hacktoberfest labels also? 😃

m1ome:
@reichert621 also it would be awesome if you assign this on to me, and i will try done this ASAP

reichert621:
sure! still trying to sort things out in this discussion (#60), but in the meantime I'd love to see what your proposed updates would look like :)

reichert621:
hey @m1ome! any updates on this?

flmel:
Me

PR #396, opened 2020-11-12
Tests factories and ex machina

flmel:
### Description
**work in progress**
Added ex_machina in order to DRY tests did rework the accounts tests
### Issue
#316 and #60 

### Screenshots

## Checklist

- [x] Ran `mix format`


test/chat_api/accounts_test.exs
```diff
@@ -1,33 +1,35 @@
...
 
   describe "accounts" do
     alias ChatApi.Accounts.Account
 
-    @valid_attrs %{company_name: "some company_name"}
-    @update_attrs %{company_name: "some updated company_name"}
-    @invalid_attrs %{company_name: nil}
+    @valid_attrs params_for(:account)
+    @update_attrs params_for(:account, company_name: "updated company name")
+    @invalid_attrs params_for(:invalid_account)
```
reichert621, line 12: would this work if we did `params_for(:account, company_name: nil)`?

I'm not sure if we actually need to define fixtures for every "invalid" type 🤔

reichert621:
looks reasonable so far @flmel! thanks :)

test/chat_api/accounts_test.exs
flmel: invalid_attrs params_for(:account, "")  would work ofcourse nil value would skip the field

reichert621:
looks great so far @flmel! it's quite long so it might take me some to review all of it 😅 but looking forward to merging your updates 🎉

flmel:
> looks great so far @flmel! it's quite long so it might take me some to review all of it 😅 but looking forward to merging your updates 🎉

the only issue would be is that it blocks further backend tasks since adding more tests would cause catch22

reichert621:
yup good point -- sorry about that! i'll make sure we prioritize this before merging any new backend tests 👍

reichert621 (commented):
amazing work on this @flmel! 🚀  🚀  🚀 

simplifies tests, and also speeds them up 😄 (i think it went from ~30s before to now around ~10s on my laptop!)

Issue #388, opened 2020-11-11
Add "Timezone" data to initial Slack conversation thread


reichert621:
**Problem**
This is a useful proxy for knowing where the user is located. It should be easy to just add this to the initial message in our Slack conversation threads.

**Solution**
Add "Timezone" to initial Slack conversation thread if it exists. If it doesn't exist, just display "N/A"

The code will go here: https://github.com/papercups-io/papercups/blob/master/lib/chat_api/slack.ex#L278-L300

**Testing**
You may need to set up our Slack integration locally to test this.

Here are the docs: https://github.com/papercups-io/papercups/wiki/Setting-up-Slack

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

kamilpikula:
hi @reichert621, I think I'll add this feature, you can assign to me. :)

reichert621:
sounds good, thanks @kamilpikula!

cheeseblubber:
@reichert621 closing now since it looks merged feel free to reopen

PR #398, opened 2020-11-14
feat: Add "Timezone" data to initial Slack conversation thread

kamilpikula:
### Problem

This is a useful proxy for knowing where the user is located. It should be easy to just add this to the initial message in our Slack conversation threads.
#388 

### Solution

Add "Timezone" to initial Slack conversation thread if it exists. If it doesn't exist, just display "N/A"
I added the `time_zone` field to function that returns payload and I improved the tests.

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/test/chat_api/slack_test.exs
+++ b/test/chat_api/slack_test.exs
@@ -87,11 +87,40 @@ defmodule ChatApi.SlackTest do
       customer: customer,
       thread: thread
     } do
-      channel = thread.slack_channel
       text = "Hello world"
+      customer_email = "*Email:*\n#{customer.email}"
+      channel = thread.slack_channel
 
       assert %{
-               "blocks" => _blocks,
+               "blocks" => [
+                 %{
+                   "text" => %{
+                     "text" => ^text
+                   }
+                 },
+                 %{
+                   "fields" => [
+                     %{
+                       "text" => "*Name:*\nAnonymous User"
+                     },
+                     %{
+                       "text" => ^customer_email
+                     },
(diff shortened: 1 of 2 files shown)


test/chat_api/slack_test.exs
```diff
@@ -87,11 +87,40 @@ defmodule ChatApi.SlackTest do
...
+                       "text" => "*URL:*\nN/A"
+                     },
+                     %{
+                       "text" => "*Browser:*\nN/A"
+                     },
+                     %{
+                       "text" => "*OS:*\nN/A"
+                     },
+                     %{
+                       "text" => "*Timezone:*\nN/A"
```
reichert621, line 119: nice! can we also add a test where these values actually exist on the customer?

(you can create a test customer with `customer_fixture(account, %{name: "...", time_zone: "...", etc: "etc"})`)

reichert621 (commented):
thanks for tackling this @kamilpikula! just one small request :)

test/chat_api/slack_test.exs
kamilpikula: @reichert621 do you think that I should write another function or improve the one above with actual values (in `get_message_payload/2 returns payload for initial slack thread` test)?

reichert621: both would be nice! that way we test that the fallback values (e.g. "N/A") are working properly :)

reichert621:
merging for now -- can add more tests later 👍

PR #399, opened 2020-11-14
Improve query conversation by account query

### LINKED ISSUE #391: Optimize the query in `ChatApi.Conversations.list_conversations_by_account`
**Problem**
This query could be made more efficient. We currently fetch all conversations ordered by `:inserted_at`, while in the UI we display them based on how recently a message was sent within it (i.e. `message.sent_at`).

**Solution**
Optimize this query to order the results by the most recently sent message associated with the conversation.

It might also be helpful to write a query to fetch all conversations with only the most recent message included along with it, rather than ALL associated messages. (Our current query won't scale...)

Method: https://github.com/papercups-io/papercups/blob/e4fb1b82d6fb4d388b4effc030efbd01c500b159/lib/chat_api/conversations.ex#L33-L41

**Testing**
Please write test cases in `/test/chat_api/conversations_test.exs`

Also please verify that this doesn't break how conversations are loaded in the UI!

WLSF:
### Description

- Add new module for Query on Conversations, to isolate the pure-functions queries from impure(functions that will perform database hits).
- Add new lateral join to retrieve(preload) only one message(most recent) by conversation.
- Improve the conversations test to validate that is been ran using a conversation that has more than three messages within.
- Add test to check if the filtered conversation has only one message.

### Issue

Fix #391 

### Screenshots

N/A

## Checklist

- [x] Ran `mix format`


lib/chat_api/conversations/query.ex
```diff
@@ -0,0 +1,43 @@
+defmodule ChatApi.Conversations.Query do
+  import Ecto.Query, warn: false
+
+  alias ChatApi.Conversations.Conversation
+  alias ChatApi.Messages.Message
+
+  def query_by_account(account_id, attrs) do
+    Conversation
+    |> join(
+      :left_lateral,
```
WLSF, line 10: Not sure if doing a left lateral here would be the most appropriated way, but I didn't manage to find another way to achieve the `subquery` to retrieve only one message(most recent) by conversations.

Any suggestions?

reichert621 (changes requested):
hey @WLSF! this looks great so far :) but one thing to notice from https://github.com/papercups-io/papercups/issues/391 is that we still want the query to return ALL the messages for now, so it doesn't break the existing behavior

eventually, i think it'll be much better if we only fetch the most recent message upfront, and only load the message history when the conversation is selected in the UI... but for now, this just gets a little closer to that :) 

(so maybe the method you have here can be put in a `list_conversations_by_account_v2` method for now, but we'll still need `list_conversations_by_account` to return all the messages along with the conversations to avoid breaking anything)

lib/chat_api/conversations/query.ex
reichert621: seems reasonable! I'm not aware of any better way to do it :P

lib/chat_api/conversations/query.ex
```diff
@@ -0,0 +1,53 @@
+defmodule ChatApi.Conversations.Query do
```
reichert621, line 1: hmm personally I feel like this abstraction isn't necessary at this point... can we keep everything in the `Conversations` context for now?

(the reason being: if I'm looking at the methods performing these queries, it's hard to know exactly what's going on if I just see `Query.by_customer`... also, it's not like we're using these queries in multiple places yet, so seems unnecessary 🤷 )

WLSF: Ok!

test/chat_api/conversations_test.exs
```diff
@@ -41,11 +41,39 @@ defmodule ChatApi.ConversationsTest do
...
+      for _n <- 1..3,
+          do:
+            message_fixture(account, secondary_conversation, %{sent_at: ~N[2020-11-14 20:50:26]})
+
+      conversations = Conversations.list_conversations_by_account_v2(account.id, %{})
+      assert length(conversations) == 2
+
+      assert [%{messages: messages}, %{messages: messages2}] = conversations
+      assert length(messages) == 1
+      assert length(messages2) == 1
```
reichert621, line 74: can we also test that the message included here is only the most recent one?

reichert621 (commented):
hey @WLSF thanks for fixing this up! 😄 

just one last request -- can we add a few more test cases to make sure your updates work as expected? (for the `list_conversations_by_account` method)

this method is used quite often so I want to make sure it's working as expected before merging 👍

test/chat_api/conversations_test.exs
WLSF: Of course, I'll do my best here.

test/chat_api/conversations_test.exs
```diff
@@ -41,11 +41,48 @@ defmodule ChatApi.ConversationsTest do
...
 
+      for _n <- 1..10,
+          do: message_fixture(account, conversation, %{sent_at: ~N[2020-11-14 20:50:26]})
+
       result_ids = Enum.map(Conversations.list_conversations_by_account(account.id), & &1.id)
 
       assert result_ids == [conversation.id]
     end
 
+    test "list_conversations_by_account_v2/2 returns all conversations with one message(most recent)",
```
reichert621, line 52: @WLSF looks like this test is failing?

WLSF: Yes, I was checking the reason... because it wasnt failing on the localhost...

WLSF: @reichert621 I've removed all the code about the list_conversations_v2, so I can work better on that in the background, and then we can back to it later when we found use for it to have a list of conversations with the top 10 messages, or something like that...

reichert621:
@WLSF I pulled your branch and did some testing, found that using `message.inserted_at` rather than `sent_at` is probably better since we still need to backfill a bunch of messages with the `sent_at` field (it's not currently required, but we added it to make it easier to measure how long it takes to save a message after it's sent in from the client/UI)

also just want to do a bit more testing since this method is so frequently used, but looks good so far!

WLSF:
@reichert621 I see, that's a good point, I wasn't aware of the differences between `inserted_at` vs `sent_at`...

reichert621:
no worries @WLSF -- it was my fault!

reichert621 (approved):
👍 seems to work!

PR #400, opened 2020-11-16
Auto-assign first agent to reply to a conversation

reichert621:
### Description

Auto-assigns the first agent to reply to a conversation

### Issue

Fixes: https://github.com/papercups-io/papercups/issues/149

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api/conversations.ex
+++ b/lib/chat_api/conversations.ex
@@ -206,16 +206,18 @@ defmodule ChatApi.Conversations do
     Conversation.changeset(conversation, attrs)
   end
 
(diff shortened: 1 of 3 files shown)


lib/chat_api/conversations.ex
```diff
@@ -113,18 +95,6 @@ defmodule ChatApi.Conversations do
   end
 
   @spec create_conversation(map()) :: {:ok, Conversation.t()} | {:error, Ecto.Changeset.t()}
-  @doc """
```
reichert621, line 116: (removed these because they feel redundant now that we have typespecs)

PR #425, opened 2020-11-30
Always show current live url of user in sidebar and  live Sessions Page

arjenrschat:
When the user is live, you still want to be able to got to the page the user is on:

### Screenshots
Current url in sidebar
image

Current url at the top of the live sessions page
image


reichert621:
thanks @arjenrschat! I like this idea, but I probably want to tweak the design a bit... let me think about it some more

reichert621:
fixed in https://github.com/papercups-io/papercups/commit/021416afb3196b51276978ef2361e49bb4892fe2

PR #426, opened 2020-11-30
Conversations list improvements

arjenrschat:
### Description

Added formatRelativeTimeShort() util
SenderAvatar can be variable size
Use short relative time in conversations list
Show agent avatar next to last message

### Screenshots
image

image


reichert621:
I like some of these ideas, but I want to clean up the design a bit first... I'll leave this open in the meantime :)

arjenrschat:
This update gives all status indicators it own location:
Unread message, who sent it, prioritized conversation, is the user online now

image

PR #427, opened 2020-11-30
Chat messages improvements

arjenrschat:
Show time next to the chat message when it sent more than an hour later than the previous message

Only show the sender avatar once per chat messages group
Fixed scrolling bar display for chat messages window
Removed ! from chat message placeholder
Doubled the height of the textarea of the chat reply message, to improve the hit area.

### Screenshots
image


reichert621:
i like these ideas! can we show the avatar for the last message rather than the first in the group though? i'm not sure which is better...

arjenrschat:
I don't know what I like better, but this works as well.


image

Issue #413, opened 2020-11-23
Hide widget outside of working hours


cheeseblubber:
**Problem**
As a company it is a bad experience if a customer messages me and there is no response

**Solution**
As a user I want to be able to hide widget outside of working hours. I want to have a boolean flag is set.

You need to modify:
https://github.com/papercups-io/chat-widget

- [ ] Widget should have a boolean field with the flag hideWidgetOutsideOfWorkingHours (Or whatever reasonable name it is) 
- [ ] When a user sets the boolean to true the widget should be hidden when it is outside of working hours
- [ ] Dashboard should have a button that toggles this on and off
https://github.com/papercups-io/papercups - dashboard change to persist the data

Edge cases:
* If the boolean is set before the user has the working hours set it should show a console log but should not hide the widget
* When a user has finished 

**Testing**
A user should be able to set the flag a minute in the future and when they refresh the flag a minute later it should hide the widget

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

cheeseblubber:
Happy to merge in just the widget side changes and dashboard changes separately if making both is too intimidating

reichert621:
(similar to https://github.com/papercups-io/papercups/issues/243)

reichert621:
(gonna assign @fkotsian for now, i can always pick this up if he's too busy though)

PR #430, opened 2020-12-01
[chat_api, dashboard] Account user can flip 'hide_widget_outside_working_hours' switch

fkotsian:
### Description

Adds `hide_widget_outside_working_hours` boolean to Accounts, and allows toggling it from AccountOverview.tsx

### Issue

Part of https://github.com/papercups-io/papercups/issues/413

### Screenshots

hide_working_hrs

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/assets/src/components/account/AccountOverview.tsx
+++ b/assets/src/components/account/AccountOverview.tsx
@@ -3,6 +3,7 @@ import {Box, Flex} from 'theme-ui';
 import {
   notification,
   Button,
+  Checkbox,
   Divider,
   Input,
   Paragraph,
@@ -13,7 +14,7 @@ import Spinner from '../Spinner';
 import AccountUsersTable from './AccountUsersTable';
 import DisabledUsersTable from './DisabledUsersTable';
 import WorkingHoursSelector from './WorkingHoursSelector';
-import {WorkingHours} from './support';
+import {WidgetSettings, WorkingHours} from './support';
 import * as API from '../../api';
 import {User} from '../../types';
 import {BASE_URL} from '../../config';
@@ -29,6 +30,17 @@ type State = {
   isLoading: boolean;
   isEditing: boolean;
   isRefreshing: boolean;
+  hideWidgetOutsideWorkingHours: boolean;
+};
+type Account = {
+  id: string;
+  company_name: string;
+  time_zone: string;
+  subscription_plan: string;
+  hide_widget_outside_working_hours: boolean;
+  users: Array<User>;
+  widget_settings: Array<WidgetSettings>;
+  working_hours: Array<WorkingHours>;
 };
 
 class AccountOverview extends React.Component<Props, State> {
@@ -42,6 +54,7 @@ class AccountOverview extends React.Component<Props, State> {
     isLoading: true,
     isEditing: false,
     isRefreshing: false,
+    hideWidgetOutsideWorkingHours: false,
   };
 
   async componentDidMount() {
@@ -53,12 +66,19 @@ class AccountOverview extends React.Component<Props, State> {
     this.setState({currentUser, isLoading: false});
   }
 
+  setAccountState = (account: Account) => {
+    this.setState({
+      account: account,
+      companyName: account.company_name,
+      hideWidgetOutsideWorkingHours: account.hide_widget_outside_working_hours,
+    })
+  }
+
   fetchLatestAccountInfo = async () => {
     const account = await API.fetchAccountInfo();
-    const {company_name: companyName} = account;
     logger.debug('Account info:', account);
 
-    this.setState({account, companyName});
+    this.setAccountState(account);
   };
 
   hasAdminRole = () => {
@@ -118,25 +138,40 @@ class AccountOverview extends React.Component<Props, State> {
     company_name?: string;
     time_zone?: string;
     working_hours?: Array<WorkingHours>;
+    hide_widget_outside_working_hours?: boolean;
   }) => {
     return API.updateAccountInfo(updates)
       .then((account) => {
         logger.debug('Successfully updated company name!', account);
 
-        this.setState({account, isEditing: false});
+        this.setAccountState(account);
       })
       .catch((err) => {
         logger.error('Failed to update company name!', err);
 
         return this.fetchLatestAccountInfo();
       })
-      .then(() => this.setState({isEditing: false}));
+      .finally(() => {
+        this.setState({isEditing: false});
+      })
   };
 
   handleUpdateCompanyName = () => {
     const {companyName} = this.state;
 
     return this.handleUpdate({company_name: companyName});
+  };
+
+  handleUpdateHideWidgetOutsideWorkingHours = () => {
+    this.setState({isRefreshing: true});
+
+    const {hideWidgetOutsideWorkingHours} = this.state;
+
+    return this.handleUpdate({
+      hide_widget_outside_working_hours: !hideWidgetOutsideWorkingHours
+    }).finally(() => {
+      this.setState({isRefreshing: false})
+    });
   };
 
   handleDisableUser = async (user: User) => {
@@ -200,6 +235,7 @@ class AccountOverview extends React.Component<Props, State> {
       isLoading,
       isEditing,
       isRefreshing,
+      hideWidgetOutsideWorkingHours,
     } = this.state;
 
     if (isLoading) {
@@ -288,6 +324,14 @@ class AccountOverview extends React.Component<Props, State> {
             onSave={this.handleUpdate}
           />
         </Box>
+
+        <Checkbox
+          checked={hideWidgetOutsideWorkingHours}
+          onChange={this.handleUpdateHideWidgetOutsideWorkingHours}
+          disabled={isRefreshing}
+        >
+          Hide chat window outside of working hours
+        </Checkbox>
         <Divider />
 
         {isAdmin && (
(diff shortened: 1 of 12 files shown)


priv/repo/migrations/20201201051531_add_hide_widget_outside_working_hours_to_accounts.exs
```diff
@@ -0,0 +1,9 @@
+defmodule ChatApi.Repo.Migrations.AddHideWidgetOutsideWorkingHoursToAccounts do
+  use Ecto.Migration
+
+  def change do
+    alter table(:accounts) do
+      add :hide_widget_outside_working_hours, :boolean
```
fkotsian, line 6: would you guys want this on Account or WidgetSettings? I went with Account bc it's not a JS-config setting, but can totally change

assets/src/components/account/AccountOverview.tsx
```diff
@@ -288,6 +324,14 @@ class AccountOverview extends React.Component<Props, State> {
...
           />
         </Box>
+
+        <Checkbox
+          checked={hideWidgetOutsideWorkingHours}
+          onChange={this.handleUpdateHideWidgetOutsideWorkingHours}
+          disabled={isRefreshing}
+        >
+          Hide chat window outside of working hours
+        </Checkbox>
```
fkotsian, line 334: Any preference on where the button lives? North/south of the Edit button? Went with quite literal text as well

assets/src/components/account/AccountOverview.tsx
```diff
@@ -53,12 +66,19 @@ class AccountOverview extends React.Component<Props, State> {
     this.setState({currentUser, isLoading: false});
   }
 
+  setAccountState = (account: Account) => {
+    this.setState({
+      account: account,
+      companyName: account.company_name,
+      hideWidgetOutsideWorkingHours: account.hide_widget_outside_working_hours,
```
fkotsian, line 73: I added this helper as we're now setting derived Account state in 2 places, lemme know if you don't like. I also don't love snake_to_camelCase, am thinking of some way to gloss that over. I rather like the destructuring `{company_name: companyName}` that was going on L58 as well, wonder if there's some way to use that

assets/src/components/account/AccountOverview.tsx
```diff
@@ -29,6 +30,17 @@ type State = {
   isLoading: boolean;
   isEditing: boolean;
   isRefreshing: boolean;
+  hideWidgetOutsideWorkingHours: boolean;
+};
+type Account = {
```
fkotsian, line 35: just a thought, could also be AccountShow if we're interested in telling the difference between Account#basic.json and Account#account.json (could def also be just `?` optional keys)

fkotsian (commented):
Just some thoughts & questions for y'all

assets/src/components/account/AccountOverview.tsx
```diff
@@ -29,6 +30,17 @@ type State = {
...
+  id: string;
+  company_name: string;
+  time_zone: string;
+  subscription_plan: string;
+  hide_widget_outside_working_hours: boolean;
+  users: Array<User>;
+  widget_settings: Array<WidgetSettings>;
```
reichert621, line 42: i think this should just be `widget_settings: WidgetSettings`, sorry for the confusing naming 😅

assets/src/components/account/AccountOverview.tsx
reichert621: i think it's fine to have it here! though i'd maybe remove the word "window", i think it still makes enough sense without it... what do you think?

assets/src/components/account/support.ts
```diff
@@ -15,6 +15,17 @@ export type WorkingHours = {
   end_minute: number;
 };
 
+export type WidgetSettings = {
```
reichert621, line 18: (as noted above, let's move this to `assets/src/types.ts`)

priv/repo/migrations/20201201051531_add_hide_widget_outside_working_hours_to_accounts.exs
reichert621: hmm the more i think about it the more i think it makes sense to have this on WidgetSettings... mainly because from an API design perspective i'd expect to see this field in a `GET /api/widget_settings` rather than a `GET /api/accounts` situation... what do you think?

(if we move it to WidgetSettings, i'd rename it to just `hide_outside_working_hours`)

reichert621 (commented):
thanks again for this @fkotsian! just a few comments :)

assets/src/components/account/AccountOverview.tsx
fkotsian: aha! sounds good 👍

assets/src/components/account/AccountOverview.tsx
fkotsian: makes sense to me! I like the elegant naming of "hideOutsideWorkingHours" as well

priv/repo/migrations/20201201051531_add_hide_widget_outside_working_hours_to_accounts.exs
fkotsian: makes sense to me! 👍

assets/src/components/account/AccountOverview.tsx
fkotsian: done!

assets/src/components/account/AccountOverview.tsx
```diff
@@ -139,6 +160,21 @@ class AccountOverview extends React.Component<Props, State> {
     return this.handleUpdate({company_name: companyName});
   };
 
+  handleUpdateHideWidgetOutsideWorkingHours = () => {
+    this.setState({isRefreshing: true});
+
+    const widgetSettingsToUpdate = {
+      id: this.state.account.widget_settings.id,
+      hide_outside_working_hours: !this.state.hideWidgetOutsideWorkingHours,
```
fkotsian, line 159: moved this to an object to reflect the desired nested association

assets/src/components/account/AccountOverview.tsx
```diff
@@ -118,19 +127,22 @@ class AccountOverview extends React.Component<Props, State> {
     company_name?: string;
     time_zone?: string;
     working_hours?: Array<WorkingHours>;
+    widget_settings?: UpdateWidgetSettings;
```
fkotsian, line 130: Phoenix notes it'll create an extra association if the nested entity doesn't possess an ID, so I added a lil guard here in the form of an UpdateType... not sure how much we're concerned about it really, more just thinking through how to protect from something like this in general.

I think a type is elegant, so is probably an update-specific changeset on the BE (tho that will show an error to the user, not amazing really). I dunno, this seems pretty good I guess. 

I definitely don't intend to start typing each type of CRUD for no reason, this just seemed like a reasonable place to do it

assets/src/components/account/support.ts
```diff
@@ -9,12 +9,6 @@ type Timezone = {
   tzCode: string;
 };
 
-export type WorkingHours = {
-  day: string;
-  start_minute: number;
-  end_minute: number;
-};
-
```
fkotsian, line 17: moved types

lib/chat_api/accounts/account.ex
```diff
@@ -45,6 +45,7 @@ defmodule ChatApi.Accounts.Account do
       :stripe_default_payment_method_id
     ])
     |> cast_embed(:working_hours, with: &working_hours_changeset/2)
+    |> cast_assoc(:widget_settings, required: false, with: &WidgetSetting.changeset/2)
```
fkotsian, line 48: this is where the magic happens

lib/chat_api/widget_settings/widget_setting.ex
```diff
@@ -18,6 +18,8 @@ defmodule ChatApi.WidgetSettings.WidgetSetting do
     field :pathname, :string
     field :last_seen_at, :utc_datetime
 
+    field(:hide_outside_working_hours, :boolean)
```
fkotsian, line 21: new field

fkotsian (commented):
@reichert621 updated for ya! did some fun stuff with an UpdateType to guard against creating extra WidgetSettings, not super worried about it but curious what you think about it

fkotsian:
@reichert621 Good to merge this guy in?

reichert621:
@fkotsian looks good but I want to make sure the chat widget updates are in place first (otherwise toggling the checkbox you added won't do anything)... maybe we can just comment out the UI part of this PR and merge in all the other logic for now until that's ready?

assets/src/components/account/AccountOverview.tsx
reichert621: hmm I wonder if it would be simpler to just use the `API.updateWidgetSettings` to handle the `widget_settings` update (since that makes sure we only update the correct `widget_settings` record based on the auth token)?

then we could just handle calling that in the `handleUpdateHideWidgetOutsideWorkingHours` without worrying about any of the other account-related fields 🤷 what do you think?

fkotsian:
Sure sounds good! I should have a draft of the widget side up tonight, barring that "what if the user's still active" edge case

PS, do we have some sort of "last activity" listener? I figure that could be an EZ/cool way to check if we should hide the chat window (afterWorkingHours && !activeFor10Minutes or sth?)

reichert621:
i think as a first pass we should only run the logic for showing/hiding the widget based on work hours when the page/widget initially loads (so it would only disappear when they reload the page after work hours are over) -- what do you think?

fkotsian:
Lol, that'd certainly be a lot easier. I just saw this message now so I put together the skeleton of a lastSeenAt timer -- lmk what you think, can certainly rip it out if need be. It uses (`messages:seen`, `messages:sent`) to update a lastSeenAt timer

https://github.com/papercups-io/chat-widget/pull/69

Issue #90, opened 2020-08-06
Redeploy shouldn’t trigger everyone to relogin


cheeseblubber:
Currently for the hosted version every deploy triggers a re login

cheeseblubber:
@reichert621 is this still an issue?

reichert621:
yes

PR #440, opened 2020-12-02
Persist auth sessions with Postgres

reichert621:
### Description

Testing out using Postgres for persistent auth sessions so we don't need to add a new service (Redis)

### Issue

Fixes https://github.com/papercups-io/papercups/issues/90

## Checklist

- [x] Ran `mix format`


lib/chat_api/auth/pow_session.ex
```diff
@@ -0,0 +1,13 @@
+defmodule ChatApi.Auth.PowSession do
+  use Ecto.Schema
+
+  @primary_key false
+  schema "pow_sessions" do
```
reichert621, line 5: (autogenerated by the `pow_postgres_store` library)

Issue #443, opened 2020-12-03
Invite url point to wrong port


arjenrschat:
On a local dev install the invite url points to the wrong port, 4000 instead of 3000

**To Reproduce**
Generate an invite url

**Screenshots**

<img width="752" alt="image" src="https://user-images.githubusercontent.com/25150192/101043750-ee97a880-357e-11eb-80f0-d1e68812bfd4.png">

Issue #444, opened 2020-12-03
On hosted version limit usage based on plan


cheeseblubber:
Debounce email validation support - when an user enters their email
story time page view
number of seats
private notes
email channel
30 day message retention

PR #447, opened 2020-12-04
Limit number of users to account based on subscription plan

reichert621:
### Description

This PR prevents accounts from exceeding the user limit of their subscription plan.

(We may need to grandfather in a few accounts that have already exceeded this limit)

### Issue

https://github.com/papercups-io/papercups/issues/444

Fixes: https://github.com/papercups-io/papercups/issues/443

### Screenshots

When new user tries to register, but the account has already reached its limit:
Screen Shot 2020-12-04 at 2 47 08 PM

When an admin users attempts to generate a new invite link after the account has reached its limit:
Screen Shot 2020-12-04 at 2 47 28 PM

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api/accounts.ex
+++ b/lib/chat_api/accounts.ex
@@ -9,6 +9,8 @@ defmodule ChatApi.Accounts do
   alias ChatApi.Accounts.Account
   alias ChatApi.Users.User
 
+  @starter_plan_max_users 5
+
   @spec list_accounts() :: [Account.t()]
   @doc """
   Returns the list of accounts.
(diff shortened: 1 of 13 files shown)


cheeseblubber:
For the admin user can we change the pop up text to: "You've hit the limit for the free tier try our premium plan for 30 days!" with a link to our pricing page

reichert621:
@cheeseblubber

Screen Shot 2020-12-04 at 3 13 34 PM

cheeseblubber:
Thought about this a little bit more can we update "Failed to generate user invitation!" to "Please upgrade to add more users?"

lib/chat_api/accounts.ex
```diff
@@ -7,6 +7,9 @@ defmodule ChatApi.Accounts do
   alias ChatApi.Repo
 
   alias ChatApi.Accounts.Account
+  alias ChatApi.Users.User
+
+  @starter_plan_max_users 5
```
reichert621, line 12: (i think this should actually be `2`)

Issue #457, opened 2020-12-08
As a CS agent I want to share specific conversation with my team


cheeseblubber:
**Problem**
I want to be able to get a link to share with my teammates for specific conversations 

**Solution**
I want to click/copy a link on the context side bar and share that to my team members for that conversation

**Additional context**
Only your teammates can get access to this conversation.

cheeseblubber:
@jamesefhawkins Would the goal of this to be to have the teammate reply to it or is the goal to share a transcript of the conversation?

fuziontech:
I think both. The let's say I spot a question in a conversation that I don't know anything about and want to pass it off to someone who does. It would be super helpful to have a deep link that I could share with them that would take them right to the specific conversation. You do this already in the slack integration including the uuid in the link.

I think the ideal case here is when you click on a conversation you update the url bar with a url that includes uuid of the thread so that you can easily share it with someone else.

reichert621:
@fuziontech should be done in #462 but feel free to reopen if i missed something!

PR #462, opened 2020-12-10
Add ability to share conversation transcript with link

reichert621:
### Description

Adds the basic ability to share a conversation transcript with a link that expires after 24 hours.

### Issue

Sets up a v1 of https://github.com/papercups-io/papercups/issues/457

### Screenshots

share-conversation-v1

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api_web/controllers/conversation_controller.ex
+++ b/lib/chat_api_web/controllers/conversation_controller.ex
@@ -75,6 +75,37 @@ defmodule ChatApiWeb.ConversationController do
     render(conn, "index.json", conversations: conversations)
   end
 
+  @spec share(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def share(conn, %{"conversation_id" => conversation_id}) do
+    with %{account_id: account_id} <- conn.assigns.current_user,
+         %{customer_id: customer_id} <- Conversations.get_conversation!(conversation_id) do
+      token = Phoenix.Token.sign(ChatApiWeb.Endpoint, conversation_id, {account_id, customer_id})
+
+      json(conn, %{data: %{ok: true, token: token}})
+    end
+  end
+
+  @spec shared(Plug.Conn.t(), map()) :: Plug.Conn.t()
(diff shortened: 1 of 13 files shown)


lib/chat_api_web/controllers/conversation_controller.ex
```diff
@@ -75,6 +75,37 @@ defmodule ChatApiWeb.ConversationController do
     render(conn, "index.json", conversations: conversations)
   end
 
+  @spec share(Plug.Conn.t(), map()) :: Plug.Conn.t()
```
reichert621, line 78: (still need to add tests for these endpoints)

lib/chat_api_web/controllers/conversation_controller.ex
```diff
@@ -75,6 +75,37 @@ defmodule ChatApiWeb.ConversationController do
...
+         %{customer_id: customer_id} <- Conversations.get_conversation!(conversation_id) do
+      token = Phoenix.Token.sign(ChatApiWeb.Endpoint, conversation_id, {account_id, customer_id})
+
+      json(conn, %{data: %{ok: true, token: token}})
+    end
+  end
+
+  @spec shared(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def shared(conn, %{"conversation_id" => conversation_id, "token" => token}) do
+    case Phoenix.Token.verify(ChatApiWeb.Endpoint, conversation_id, token, max_age: 86400) do
```
reichert621, line 90: (maybe make it clearer that `86400` is 24 hours)

cheeseblubber:
Would it have been easier if we have a route for the conversation id and then we wouldn't need the token and just rely on our auth? This way if the person receiving the link can navigate to the conversation to reply. Because right now if a user that received the link wants to reply to that conversation this still wouldn't solve that issue

reichert621:
ah i guess i misunderstood the ticket 😑

fuziontech:
This is perfect. Thanks!

Issue #433, opened 2020-12-02
Customer Notes v1


reichert621:
**Problem**
As a user, I should be able to add notes to a customer.

At a high level, this will involve the following tasks:
- Creating a new table/model in the db to store the notes
  - For reference, here's Intercom's Notes API: https://developers.intercom.com/intercom-api-reference/v0/reference#note-model
- Creating the APIs to create/read/update/delete a note
  - And ideally some tests! (Phoenix will autogenerate some tests)
- Updating the conversation sidebar UI to display notes and allow the user to add/edit/delete notes

For reference, here's how notes look in Intercom's UI:
<img width="1014" alt="intercom-sidebar-notes" src="https://user-images.githubusercontent.com/5264279/100820971-6483fe80-341d-11eb-813f-174efc7f2c26.png">

(Note that Intercom calls them "leads"... we should still call them "customers" for now to be consistent with the rest of our app)

**Solution**

TODO: outline the above in more details:
- The fields we should include on the Notes model/table
- The API endpoints we'll set up for this feature
- A proposal for what the UX should be like
  - (e.g. do we create/edit the notes inline vs in a modal?)

(For reference, here's how we broke down the tasks for our Reporting dashboard: https://github.com/papercups-io/papercups/issues/252)

(Also if you're new to Elixir/Phoenix, here's an example from a while ago showing how we set up one of our Conversations APIs: https://github.com/papercups-io/papercups/blob/master/docs/examples/add_conversations_api.md)

reichert621:
@fkotsian mind if I assign this to you? 😄

fkotsian:
3 pieces:

> - Creating a new table/model in the db to store the notes

Note
- author_id: int
- account_id: uuid
- customer_id: uuid
- body: text
- timestamps()

> - Creating the APIs to create/read/update/delete a note
>  - And ideally some tests! (Phoenix will autogenerate some tests)

NoteController

def create(conn, %{"note" => note_params}) do  # params = %{"body" => body}
  # require body
  # author_id = current_user, account_id = current_user.account
end

def index(conn, params) do
  # with %{account_id: account_id} <- conn.assigns.current_user do
  #   notes = Notes.list_notes(account_id)
  # render(conn, "index.json", notes: notes)
end

# NB: I kinda think we don't need this endpoint? Fetch 1 specific note? Notes will always be fetched in batches AFAICT
def show(conn, %{id => id}) do
  # Notes.get_note(id)
[...]

PR #464, opened 2020-12-10
[dashboard] customer notes backend

fkotsian:
#### Schema
```
Note:
- id: int
- author_id: int
- account_id: uuid
- customer_id: uuid
- body: text
- timestamps()
```

### Issue

Part of https://github.com/papercups-io/papercups/issues/433

### Screenshots

N/A

## Checklist

- [X] Ran `mix format`

#### Current UX:
N/A

### DIFF
--- a/lib/chat_api/notes/note.ex
+++ b/lib/chat_api/notes/note.ex
@@ -2,13 +2,14 @@ defmodule ChatApi.Notes.Note do
   use Ecto.Schema
   import Ecto.Changeset
 
+  alias ChatApi.Accounts.Account
+  alias ChatApi.Customers.Customer
+  alias ChatApi.Users.User
+
   @primary_key {:id, :binary_id, autogenerate: true}
   @foreign_key_type :binary_id
   schema "notes" do
     field :body, :string
-    field :author_id, :integer
-    field :account_id, :binary_id
-    field :customer_id, :binary_id
 
     belongs_to(:account, Account)
     belongs_to(:customer, Customer)
@@ -28,6 +29,6 @@ defmodule ChatApi.Notes.Note do
   def changeset(note, attrs) do
     note
     |> cast(attrs, [:body, :author_id, :account_id, :customer_id])
-    |> validate_required([])
+    |> validate_required([:body])
   end
 end
--- a/lib/chat_api/notes/note.ex
+++ b/lib/chat_api/notes/note.ex
@@ -18,9 +18,16 @@ defmodule ChatApi.Notes.Note do
   end
 
   @doc false
+  def create_changeset(note, attrs) do
+    note
+    |> cast(attrs, [:body, :author_id, :account_id, :customer_id])
+    |> validate_required([:body, :author_id, :account_id, :customer_id])
+  end
+
+  @doc false
   def changeset(note, attrs) do
     note
-    |> cast(attrs, [:body, :author_id, :customer_id])
-    |> validate_required([:body, :author_id, :customer_id])
+    |> cast(attrs, [:body, :author_id, :account_id, :customer_id])
+    |> validate_required([])
   end
 end
--- a/lib/chat_api/notes/note.ex
+++ b/lib/chat_api/notes/note.ex
@@ -6,7 +6,7 @@ defmodule ChatApi.Notes.Note do
   @foreign_key_type :binary_id
   schema "notes" do
     field :body, :string
-    field :author_id, :binary_id
+    field :author_id, :integer
     field :account_id, :binary_id
     field :customer_id, :binary_id
 
@@ -16,7 +16,7 @@ defmodule ChatApi.Notes.Note do
   @doc false
   def changeset(note, attrs) do
     note
-    |> cast(attrs, [:body])
-    |> validate_required([:body])
+    |> cast(attrs, [:body, :author_id, :customer_id])
+    |> validate_required([:body, :author_id, :customer_id])
   end
 end
--- a/lib/chat_api/notes/note.ex
+++ b/lib/chat_api/notes/note.ex
@@ -1,0 +1,22 @@
+defmodule ChatApi.Notes.Note do
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @primary_key {:id, :binary_id, autogenerate: true}
+  @foreign_key_type :binary_id
+  schema "notes" do
+    field :body, :string
+    field :author_id, :binary_id
+    field :account_id, :binary_id
+    field :customer_id, :binary_id
+
+    timestamps()
+  end
+
+  @doc false
+  def changeset(note, attrs) do
+    note
+    |> cast(attrs, [:body])
+    |> validate_required([:body])
+  end
+end
--- a/lib/chat_api/notes/note.ex
+++ b/lib/chat_api/notes/note.ex
@@ -19,16 +19,10 @@ defmodule ChatApi.Notes.Note do
   end
 
   @doc false
-  def create_changeset(note, attrs) do
-    note
-    |> cast(attrs, [:body, :author_id, :account_id, :customer_id])
-    |> validate_required([:body, :author_id, :account_id, :customer_id])
-  end
-
-  @doc false
   def changeset(note, attrs) do
     note
     |> cast(attrs, [:body, :author_id, :account_id, :customer_id])
-    |> validate_required([:body])
+    # partial updates are allowed if required value is already populated: https://hexdocs.pm/ecto/Ecto.Changeset.html#validate_required/3
+    |> validate_required([:body, :author_id, :account_id, :customer_id])
   end
 end
--- a/lib/chat_api/notes/note.ex
+++ b/lib/chat_api/notes/note.ex
@@ -10,6 +10,10 @@ defmodule ChatApi.Notes.Note do
     field :account_id, :binary_id
     field :customer_id, :binary_id
 
+    belongs_to(:account, Account)
+    belongs_to(:customer, Customer)
+    belongs_to(:author, User, foreign_key: :author_id, references: :id, type: :integer)
+
     timestamps()
   end
 
--- a/lib/chat_api_web/controllers/note_controller.ex
+++ b/lib/chat_api_web/controllers/note_controller.ex
@@ -7,11 +7,11 @@ defmodule ChatApiWeb.NoteController do
   action_fallback ChatApiWeb.FallbackController
 
   @spec index(Plug.Conn.t(), map()) :: Plug.Conn.t()
-  def index(conn, %{"customer_id" => customer_id} = note_params) do
+  def index(conn, note_params) do
     case note_params do
-      %{"customer_id" => ^customer_id} ->
+      %{"customer_id" => customer_id} ->
         with %{account_id: account_id} <- conn.assigns.current_user do
-          notes = Notes.list_notes_for_customer(%{account_id: account_id, customer_id: customer_id})
+          notes = Notes.list_notes_for_customer([account_id: account_id, customer_id: customer_id])
           render(conn, "index.json", notes: notes)
         end
 
@@ -31,13 +31,13 @@ defmodule ChatApiWeb.NoteController do
   def create(conn, %{"note" => note_params}) do
     %{"body" => body, "customer_id" => customer_id} = note_params
 
-    with %{account_id: account_id, id: author_id} <- conn.assigns.current_user do
-      {:ok, %Note{} = note} = Notes.create_note(%{
+    with %{account_id: account_id, id: author_id} <- conn.assigns.current_user, 
+      {:ok, %Note{} = note} <- Notes.create_note(%{
          body: body,
          customer_id: customer_id,
          account_id: account_id,
          author_id: author_id
-       })
+       }) do
 
         conn
         |> put_status(:created)
--- a/lib/chat_api_web/controllers/note_controller.ex
+++ b/lib/chat_api_web/controllers/note_controller.ex
@@ -6,11 +6,13 @@ defmodule ChatApiWeb.NoteController do
 
   action_fallback ChatApiWeb.FallbackController
 
+  @spec index(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def index(conn, _params) do
     notes = Notes.list_notes()
     render(conn, "index.json", notes: notes)
   end
 
+  @spec create(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def create(conn, %{"note" => note_params}) do
     with {:ok, %Note{} = note} <- Notes.create_note(note_params) do
       conn
@@ -20,11 +22,13 @@ defmodule ChatApiWeb.NoteController do
     end
   end
 
+  @spec show(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def show(conn, %{"id" => id}) do
     note = Notes.get_note!(id)
     render(conn, "show.json", note: note)
   end
 
+  @spec update(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def update(conn, %{"id" => id, "note" => note_params}) do
     note = Notes.get_note!(id)
 
@@ -33,6 +37,7 @@ defmodule ChatApiWeb.NoteController do
     end
   end
 
+  @spec delete(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def delete(conn, %{"id" => id}) do
     note = Notes.get_note!(id)
 
--- a/lib/chat_api_web/controllers/note_controller.ex
+++ b/lib/chat_api_web/controllers/note_controller.ex
@@ -11,7 +11,7 @@ defmodule ChatApiWeb.NoteController do
     case note_params do
       %{"customer_id" => customer_id} ->
         with %{account_id: account_id} <- conn.assigns.current_user do
-          notes = Notes.list_notes_for_customer([account_id: account_id, customer_id: customer_id])
+          notes = Notes.list_notes_for_customer(account_id: account_id, customer_id: customer_id)
           render(conn, "index.json", notes: notes)
         end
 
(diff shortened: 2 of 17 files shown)


fkotsian:
@reichert621 Here's a first stab at the functionality for the CustomerNotes feature. As you said I think a lot of work is on the FE/UX we want -- thought we could approve/sync on this and then polish the UX on Friday?

#### Open Qs in this PR

- I am wondering if the account_id key should be :through customer (since customer belongs_to an :account? Or how should we fetch/pass the account in to the ConvDetailsSidebar)
- # of notes to display? (suggest: 3?)
- note 'show' UI? (suggest: bordered box, as in Intercom?)
- error state (suggest: border highlight?)
- delete ux? (suggest: 'x'?)
- edit ux? (suggest: none?)
- should call the table Notes.CustomerNote customer_notes as opp to Notes.Note (future flexibility?)

#### UXs needed:

- delete (trigger? modal? confirmation?)
- edit (suggest: no edit at this time?)
- validate/error (suggest: a simple border highlight - Papercups red? no tooltip - only poss error state is "no input"?)
- any others?

My suggestion would be to go with the most minimal experience for each of these and refine later? Whatever is best for the product of course 🙌

test/chat_api/notes_test.exs
```diff
@@ -0,0 +1,64 @@
...
+  alias ChatApi.Notes
+
+  describe "notes" do
+    alias ChatApi.Notes.Note
+
+    @valid_attrs %{body: "some body"}
+    @update_attrs %{body: "some updated body"}
+    @invalid_attrs %{body: nil}
+
+    def note_fixture(attrs \\ %{}) do
```
flmel, line 13: We have moved to factories from fixtures. Have a look on factory.ex you will need to create something like 

```
def note_factory do 
    %ChatApi.Notes.Note{ 
          account: build(:account),
          author: build(:user),
          customer: build(:customer)
     }
end
```
You will need to import ChatApi.Factory and then within your tests you will have access to ex_machinas comands
insert,insert_pair,insert_list, build, params_for, params_with_assocs for more info you can see [ex_machina docs](https://hexdocs.pm/ex_machina/readme.html) or peek into accounts_test.ex (or other tests) to see it in use.

Let me know if I can be of any help.

lib/chat_api/notes/note.ex
```diff
@@ -0,0 +1,22 @@
...
+    field :customer_id, :binary_id
+
+    timestamps()
+  end
+
+  @doc false
+  def changeset(note, attrs) do
+    note
+    |> cast(attrs, [:body, :author_id, :customer_id])
+    |> validate_required([:body, :author_id, :customer_id])
```
reichert621, line 20: don't forget the `account_id` on the two lines above!

lib/chat_api/notes/note.ex
```diff
@@ -0,0 +1,22 @@
...
+  import Ecto.Changeset
+
+  @primary_key {:id, :binary_id, autogenerate: true}
+  @foreign_key_type :binary_id
+  schema "notes" do
+    field :body, :string
+    field :author_id, :integer
+    field :account_id, :binary_id
+    field :customer_id, :binary_id
+
```
reichert621, line 12: can you set up the `belongs_to` relations here too? e.g.

```elixir
    belongs_to(:account, Account)
    belongs_to(:customer, Customer)
    belongs_to(:author, User, foreign_key: :author_id, references: :id, type: :integer)
```

priv/repo/migrations/20201210043230_create_notes.exs
```diff
@@ -0,0 +1,19 @@
+defmodule ChatApi.Repo.Migrations.CreateNotes do
+  use Ecto.Migration
+
+  def change do
+    create table(:notes, primary_key: false) do
+      add :id, :binary_id, primary_key: true
+      add :body, :text
+      add :author_id, references(:users, on_delete: :nothing, type: :id)
+      add :account_id, references(:accounts, on_delete: :nothing, type: :binary_id)
```
reichert621, line 9: let's do `on_delete: :delete_all` here -- if an account is deleted, we want all the associated notes to be deleted as well

priv/repo/migrations/20201210043230_create_notes.exs
```diff
@@ -0,0 +1,19 @@
...
+      add :customer_id, references(:customers, on_delete: :nothing, type: :binary_id)
```
reichert621, line 10: likewise here -- if a customer is deleted, all its associated notes should be deleted as well

test/chat_api/notes_test.exs
reichert621: i think `author` is fine, since it's a bit more descriptive than `user` (Intercom uses `author` as well in their notes API: https://developers.intercom.com/intercom-api-reference/reference#note-model -- not that we should blindly copy them, but i think it's a reasonable name 😄 )

lib/chat_api_web/router.ex
```diff
@@ -98,7 +98,11 @@ defmodule ChatApiWeb.Router do
     resources("/accounts", AccountController, only: [:update, :delete])
     resources("/messages", MessageController, except: [:new, :edit])
     resources("/conversations", ConversationController, except: [:new, :edit, :create])
-    resources("/customers", CustomerController, except: [:new, :edit, :create])
+
+    resources("/customers", CustomerController, except: [:new, :edit, :create]) do
+      resources("/notes", NoteController, except: [:new, :edit])
```
reichert621, line 103: i'd prefer not to nest this resource -- in general, i think it's best to keep as many resources at the top level (i.e. `/api/notes` if possible. (this is how we did it at stripe, and i think this stack overflow comment sums up my feelings nicely: https://stackoverflow.com/a/36410780)

lib/chat_api_web/controllers/note_controller.ex
```diff
@@ -0,0 +1,51 @@
...
+  use ChatApiWeb, :controller
+
+  alias ChatApi.Notes
+  alias ChatApi.Notes.Note
+
+  action_fallback ChatApiWeb.FallbackController
+
+  @spec index(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def index(conn, _params) do
+    notes = Notes.list_notes()
```
reichert621, line 11: this method currently fetches all of the notes in the entire database -- let's make sure we filter by `customer_id` and/or `account_id`

(remember, we can get the `account_id` from `conn.assigns.current_user.account_id`)

lib/chat_api_web/controllers/note_controller.ex
```diff
@@ -0,0 +1,51 @@
...
+    render(conn, "index.json", notes: notes)
+  end
+
+  @spec create(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def create(conn, %{"note" => note_params}) do
+    with {:ok, %Note{} = note} <- Notes.create_note(note_params) do
```
reichert621, line 17: let's make sure when we're creating a new `note`, we don't allow the client to specify the `account_id` or the `author_id` -- these should come directly from `conn.assigns.current_user`

lib/chat_api_web/views/note_view.ex
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def render("show.json", %{note: note}) do
+    %{data: render_one(note, NoteView, "note.json")}
+  end
+
+  def render("note.json", %{note: note}) do
+    %{
+      id: note.id,
+      body: note.body
```
reichert621, line 16: i think it's fine to include the `customer_id` and `author_id` here too, will probably help with debugging

test/chat_api_web/controllers/note_controller_test.exs
```diff
@@ -0,0 +1,88 @@
...
+        get(conn, Routes.note_path(conn, :show, note))
+      end
+    end
+  end
+
+  defp create_note(_) do
+    note = fixture(:note)
+    %{note: note}
+  end
+end
```
reichert621, line 88: let's get all the backend stuff fixed with tests passing before we worry about the UX :)

reichert621 (changes requested):
hey @fkotsian -- nice work so far! 😄 

i mainly focused on reviewing the backend code to start off... it would be great to get all the backend stuff fixed up with tests passing and then merge that in a separate PR before we worry about the UX 👍

test/chat_api/notes_test.exs
reichert621: i think `author` is fine, since it's a bit more descriptive than `user` (Intercom uses `author` as well in their notes API: https://developers.intercom.com/intercom-api-reference/reference#note-model -- not that we should blindly copy them, but i think it's a reasonable name 😄 )

lib/chat_api/notes/note.ex
reichert621: (might be nice to add the `has_many` relation to the `ChatApi.Customers.Customer` module in `customer.ex` as well)

lib/chat_api_web/controllers/note_controller.ex
fkotsian: Great, this clears up my main question

lib/chat_api_web/router.ex
fkotsian: V nice! People def have differing opinions about this

lib/chat_api/notes/note.ex
fkotsian: Nice

lib/chat_api_web/controllers/note_controller.ex
fkotsian: Nice call!

lib/chat_api_web/router.ex
fkotsian: Works for me!

test/chat_api/notes_test.exs
fkotsian: Thanks @flmel , the factories are much nicer to use than fixtures! I gave it a shot - let me know what you think?

lib/chat_api/notes/note.ex
```diff
@@ -0,0 +1,34 @@
...
+    belongs_to(:account, Account)
+    belongs_to(:customer, Customer)
+    belongs_to(:author, User, foreign_key: :author_id, references: :id, type: :integer)
+
+    timestamps()
+  end
+
+  @doc false
+  def create_changeset(note, attrs) do
```
reichert621, line 22: i think we can just use `changeset` below, this `create_changeset` shouldn't be necessary 👍 (reference: https://hexdocs.pm/ecto/Ecto.Changeset.html#validate_required/3)

> If a field is given to validate_required/3 but it has not been passed as parameter during cast/3 (i.e. it has not been changed), then validate_required/3 will check for its current value in the data. If the data contains a non-empty value for the field, then no error is added. **This allows developers to use validate_required/3 to perform partial updates. For example, on insert all fields would be required, because their default values on the data are all nil, but on update, if you don't want to change a field that has been previously set, you are not required to pass it as a parameter**, since validate_required/3 won't add an error for missing changes as long as the value in the data given to the changeset is not empty.

priv/repo/migrations/20201210043230_create_notes.exs
```diff
@@ -0,0 +1,19 @@
+defmodule ChatApi.Repo.Migrations.CreateNotes do
+  use Ecto.Migration
+
+  def change do
+    create table(:notes, primary_key: false) do
+      add :id, :binary_id, primary_key: true
+      add :body, :text
```
reichert621, line 7: sorry i missed this earlier -- let's add `null: false` to all the required fields, e.g.
```
add :body, :text, null: false
```

test/chat_api_web/controllers/note_controller_test.exs
```diff
@@ -0,0 +1,123 @@
...
+    } do
+      conn =
+        get(authed_conn, Routes.note_path(authed_conn, :index), %{
+          "note" => %{customer_id: note.customer_id}
+        })
+
+      note_ids = json_response(conn, 200)["data"] |> Enum.map(& &1["id"])
+
+      assert note_ids == [note.id]
+    end
```
reichert621, line 43: can we also add a test to see what happens when a `customer_id` param is missing from the request?

(ideally, it would return a `400` error)

test/chat_api_web/controllers/note_controller_test.exs
```diff
@@ -0,0 +1,123 @@
...
+
+      assert %{
+               "id" => created_id,
+               "customer_id" => customer_id,
+               "author_id" => agent_id
+             } = json_response(conn, 201)["data"]
+
+      # note is fetchable
+      conn = get(authed_conn, Routes.note_path(authed_conn, :show, created_id))
+      _fetched = json_response(conn, 200)["data"]
```
reichert621, line 69: just curious, why `_fetched` here instead of just `fetched`?

test/chat_api_web/controllers/note_controller_test.exs
```diff
@@ -0,0 +1,123 @@
...
+    } do
+      # update note
+      conn = put(authed_conn, Routes.note_path(authed_conn, :update, note), note: @update_attrs)
+      assert %{"id" => ^id} = json_response(conn, 200)["data"]
+
+      # fetch updated note
+      conn = get(authed_conn, Routes.note_path(conn, :show, id))
+
+      assert %{
+               "id" => _id,
```
reichert621, line 100: i think we can use `^id` here as well to make sure the id still matches above?

test/chat_api/notes_test.exs
```diff
@@ -0,0 +1,84 @@
...
+    setup do
+      account = insert(:account)
+      agent = insert(:user, account: account)
+      customer = insert(:customer, account: account)
+      note = insert(:note, body: "note_body", account: account, author: agent, customer: customer)
+
+      {:ok, agent: agent, account: account, customer: customer, note: note}
+    end
+
+    test "list_notes/1 returns all notes for the given account and customer", %{
```
reichert621, line 22: typo -- `list_notes` -> `list_notes_for_customer`

lib/chat_api_web/controllers/note_controller.ex
```diff
@@ -0,0 +1,65 @@
...
+      render(conn, "index.json", notes: notes)
+    end
+  end
+
+  @spec create(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def create(conn, %{"note" => note_params}) do
+    %{"body" => body, "customer_id" => customer_id} = note_params
+
+    with %{account_id: account_id, id: author_id} <- conn.assigns.current_user do
+      with {:ok, %Note{} = note} <-
```
reichert621, line 24: no need to nest `with` statements :) we can just do:
```elixir
with %{account_id: account_id, id: author_id} <- conn.assigns.current_user,
  {:ok, %Note{} = note} <- Notes.create_note(%{...}) do
  # etc
end
```

(some docs on `with`: https://elixirschool.com/en/lessons/basics/control-structures/#with)

lib/chat_api_web/controllers/note_controller.ex
```diff
@@ -0,0 +1,65 @@
+defmodule ChatApiWeb.NoteController do
+  use ChatApiWeb, :controller
+
+  alias ChatApi.Notes
+  alias ChatApi.Notes.Note
+
+  action_fallback ChatApiWeb.FallbackController
+
+  @spec index(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def index(conn, %{"note" => note_params}) do
```
reichert621, line 10: for the params, i think we can simplify it to just `%{"customer_id" => customer_id}` (rather than `%{"note" => "customer_id" => customer_id}}`)

then the API request would just look like
```
GET /api/notes?customer_id=xxxxx
```

test/chat_api/notes_test.exs
```diff
@@ -0,0 +1,84 @@
...
+                 author_id: agent.id
+               })
+
+      assert note.body == "some body"
+      assert note.customer_id == customer.id
+      assert note.author_id == agent.id
+      assert note.account_id == account.id
+    end
+
+    test "create_note/1 with invalid data returns error changeset" do
```
reichert621, line 59: can we also test that attempting to create a note with an empty string body also returns an error changeset?

reichert621 (changes requested):
nice! backend stuff looks great, just a few minor things and then it should be good to go 👍

lib/chat_api/notes/note.ex
fkotsian: Updated! My thought here was actually that allowing the entire changeset in update requests allows the client to change the Customer or the Account that the note displays on -- can add back if we're concerned about this!

test/chat_api_web/controllers/note_controller_test.exs
fkotsian: ooh good catch, test shoulda been an `==`

test/chat_api_web/controllers/note_controller_test.exs
fkotsian: These pins got me... I think I got 'em now. 

- pin when var already exists
- otherwise gets interpreted as reassignment

lib/chat_api_web/controllers/note_controller.ex
fkotsian: Ah: with is a context (return context), not an extraction operator (that's just the `<-` part) 💡

fkotsian: Huh, it took me a while but I think I get this..

So
```
with %{account_id: account_id, id: author_id} <- conn.assigns.current_user,
  {:ok, %Note{} = note} <- Notes.create_note(%{...}) do
```

if the {:ok, %Note{}} doesn't match (for ex, if create_note returns an {:error, Changeset}), it just falls through to outside the `with` (and in this case is handled by our FallbackController)?

fkotsian:
@reichert621 reverted all the frontend diff for ya, should be good to go!

reichert621 (approved):
woohoo looks great! thanks @fkotsian 🎉🎉🎉

Issue #454, opened 2020-12-07
[Free tier] Set up a worker to archive stale conversations after 30 days


reichert621:
**Problem**
We want to limit data retention for users in our hosted free tier. Let's start by archiving conversations that haven't been active in >30 days and are not marked as a priority (i.e. favorited/starred).

**Solution**
Create a worker with Oban ([example](https://github.com/papercups-io/papercups/blob/master/lib/workers/archive_conversations.ex)) that archives conversations where the following conditions are true:
- the account is on our free tier (e.g. `account.subscription_plan == "starter"`)
- the last message sent in the conversation was created >30 days ago (e.g. `message.inserted_at`)
- the conversation is not prioritized (e.g. `conversation.priority == "not_priority"`)

(We could name this worker `ChatApi.Workers.ArchiveFreeTierConversations`, or something like that)

_:warning: Note that we should probably rename the worker in https://github.com/papercups-io/papercups/blob/master/lib/workers/archive_conversations.ex to be something more like `ChatApi.Workers.ArchiveClosedConversations`_

**Additional context**
For context, the relations work as follows:
- `conversation` belongs to `account` (via `conversation.account_id`)
- `message` belongs to `conversation` (via `message.conversation_id`); `conversation` has many `message`s

Also related to https://github.com/papercups-io/papercups/issues/444

**Testing**
Please add some tests for the method that handles the above logic :)

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

edgarkanye:
@reichert621 I would like to take this on.

reichert621:
sure @edgarkanye! i'll assign you for now 👍

reichert621:
thanks @edgarkanye! (forgot to close this 😅 )

PR #471, opened 2020-12-14
Archive Old FreeTier Conversations

edgarkanye:
### Description

- rename  `ArchiveConversations`  worker to `ArchivedClosedConversations`
- create `ArchiveFreeTierConversations` worker
- Write query to handle the archiving of old/stale free tier conversations
- Adding the newly added worker to the cron tab under the `default` queue
- Writing tests

### Issue

#454 

### Screenshots

N/A

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api/conversations.ex
+++ b/lib/chat_api/conversations.ex
@@ -184,7 +184,8 @@ defmodule ChatApi.Conversations do
     Repo.update_all(query, set: [archived_at: DateTime.utc_now()])
   end
 
-  @spec query_free_tier_conversations_inactive_for([{:days, number | Decimal.t()}, ...]) ::
+  # TODO: I wonder if this should live somewhere else...
+  @spec query_free_tier_conversations_inactive_for([{:days, number}]) ::
           Ecto.Query.t()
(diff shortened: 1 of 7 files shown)


lib/chat_api/conversations.ex
```diff
@@ -183,6 +184,27 @@ defmodule ChatApi.Conversations do
     Repo.update_all(query, set: [archived_at: DateTime.utc_now()])
   end
 
+  @spec find_freetier_conversations_old_by([{:days, number | Decimal.t()}, ...]) :: Ecto.Query.t()
+  def find_freetier_conversations_old_by(days: days) do
```
reichert621, line 188: nitpick -- can we rename this to `query_free_tier_conversations_inactive_for(days: days)`? (`old_by` sounds kinda strange, and i think it's better to start with `query_` rather than `find_` so that it's clear that this is just returning a query and not actually hitting the db)

reichert621 (changes requested):
looks great @edgarkanye! just one small nitpick :)

reichert621 (approved):
thanks for the help on this @edgarkanye! 🙏

PR #488, opened 2020-12-22
Fixes #156 - Deprecate "shout" event

### LINKED ISSUE #156: Rename "shout" event to "message:created"
This name is just the default event used when generating new channels in Phoenix, but let's make our event names a bit more descriptive... we already have `conversation:created` and `conversation:updated` events, so let's be consistent and change `shout` to `message:created`.

xprazak2:
Deprecate "shout" event for conversation channel
in favor of "message:created".

### Description

_Please describe the changes you made._

### Issue

_Link to the issue (if available)_

### Screenshots

## Checklist

- [ ] Ran `mix format`


xprazak2:
I think the removal can be done in 3 steps:

1. Handle `message:created` and deprecate `shout` which keeps backward compatibility
2. Migrate clients to new event name
3. Remove `shout` in a future version when clients are migrated

lib/chat_api_web/channels/conversation_channel.ex
```diff
@@ -80,20 +81,33 @@ defmodule ChatApiWeb.ConversationChannel do
...
+  end
+
+  def handle_in("message:created", payload, socket) do
+    handle_in_msg("message:created", payload, socket)
+  end
+
   # It is also common to receive messages from the client and
   # broadcast to everyone in the current topic (conversation:lobby).
-  def handle_in("shout", payload, socket) do
+  defp handle_in_msg(event_name, payload, socket) do
```
reichert621, line 99: nitpick -- can we name this something like `handle_incoming_message` instead? i think it's a bit more explicit :) 

also let's move this to the bottom with the other private methods so we can fix the warning below 👍

reichert621 (changes requested):
just one small nitpick, and could we also handle this in the `notification_channel` as well?

thanks!

lib/chat_api_web/channels/conversation_channel.ex
xprazak2: I renamed the method and moved it to the bottom - naming is hard ;-)

I also made similar changes in `notification_channel`.

test/chat_api_web/channels/notification_channel_test.exs
```diff
@@ -38,6 +38,22 @@ defmodule ChatApiWeb.NotificationChannelTest do
     assert_push("shout", _msg)
   end
 
+  test "message:created broadcasts to notification:lobby", %{
```
reichert621, line 41: looks like this test is failing?

reichert621 (changes requested):

test/chat_api_web/channels/notification_channel_test.exs
xprazak2: Fixed.

PR #490, opened 2020-12-22
fixes #450 - Paginate list_customers

### LINKED ISSUE #450: Implement pagination for the `ChatApi.Customers.list_customers/1` method
**Problem**
We currently fetch ALL customer records when we call the `ChatApi.Customers.list_customers/1` method (i.e. we set no `limit` or `offset` in our query, nor do we have cursors or anything like that)

This will not scale very well as more and more data gets added.

**Solution**
To start off, let's do the following:
- Implement another version of `ChatApi.Customers.list_customers(account_id)` that paginates the response, either by using a `limit` and an `offset` in the query, or cursors
  - Here's where the method lives in the code: https://github.com/papercups-io/papercups/blob/master/lib/chat_api/customers.ex#L22
  - Possible library for `limit/offset` approach: https://github.com/drewolson/scrivener_ecto
  - Possible library for cursor approach: https://github.com/duffelhq/paginator
- (Note: I'm not sure that the cursor approach would even work in this scenario, since we want to be able to e.g. "fetch page 2 of our list of customers", which I'm not sure is possible with cursors... but I have a very minimal understanding of how to use cursors, so I may be wrong here)
- This new method should just be called something like `ChatApi.Customers.list_customers_paginated(account_id, pagination_options)`, and should only return, let's say, 50 records at a time

**Testing**
For the sake of this ticket, DO NOT use this new method in the `GET /api/customers` endpoint. We just want to set up the method itself, and write some tests showcasing how it works.

flmel:
### Description
Decided to go on with scrivener as we'll probably need to paginate other queries too.
added scrivener, with global page_size limit 50 (can be modified via the params)
created new list_customers/2 in customers context. Note that I've used map altho  scrivener also supports keyword lists as parameters. 

### Issue

fix #450 

### Screenshots

## Checklist

- [x] Ran `mix format`


test/chat_api/customers_test.exs
```diff
@@ -39,6 +39,21 @@ defmodule ChatApi.CustomersTest do
...
+
+      page = Customers.list_customers(account.id, %{})
+      page_with_params = Customers.list_customers(account.id, %{page: 2, page_size: 5})
+
+      assert page.total_entries == 10
+      assert page.total_pages == 1
+
+      assert page_with_params.page_number == 2
+      assert page_with_params.page_size == 5
+      assert page_with_params.total_pages == 2
```
reichert621, line 54: can we also add an assertion that all the `page.entries` have the correct `account_id`?

reichert621 (commented):
looks great @flmel 😄 just one small request!

lib/chat_api/customers.ex
```diff
@@ -23,6 +23,24 @@ defmodule ChatApi.Customers do
     Customer |> where(account_id: ^account_id) |> Repo.all()
   end
 
+  @spec list_customers(binary(), map()) :: Scrivener.Page.t()
+  @doc """
+  Returns an`%Scrivener.Page{}` with paginatinated customers.
```
reichert621, line 28: ah sorry just saw this -- small typo, should be:
```
Returns a `%Scrivener.Page{}` with paginated customers.
```

reichert621, line 28:
```suggestion
  Returns a `%Scrivener.Page{}` with paginated customers.
```

reichert621 (approved):
nice! 👍

Issue #486, opened 2020-12-21
Add `closed_at` field to Conversation model


reichert621:
**Problem**
We need to be able to calculate how long it takes for a conversation to be closed ("resolved").

Right now the way we determine if a conversation is closed is by looking at the `status` field, which can currently only be `"open"` or `"closed"`.

**Solution**
- Set up a database migration to add the `closed_at` field to the `conversations` table (as a `utc_datetime`)
- Update the `ChatApi.Conversations.Conversation` module appropriately (e.g. add the new field, also add it to the `changeset`)
- Update the `ChatApi.Conversations.update_conversation` method to set the `closed_at` field to the current time if a `status: "closed"` update is coming through (and if a `status: "open"` update is coming through, make sure to unset the `closed_at` field by setting it to `null`/`nil`)
  - Not sure what the conventional way of handling this is... I wonder if it's possible to do in the `changeset` method? Might ask the Elixir Slack group what the most idiomatic way is 🤔 
  - **Update:** this seems like it might be the best approach: https://elixirforum.com/t/adding-calculated-field-on-save/14745/3

**Describe alternatives you've considered**
It might make sense to eventually have separate, explicit API endpoints for opening/closing a conversation (rather than just handling the logic in the `update` method), but for now the above solution should require less upfront refactoring.

**Additional context**
This will be necessary to handle https://github.com/papercups-io/papercups/issues/484

**Testing**
We'll want to write tests to make sure that the `closed_at` field is set/unset appropriately based on the `status` updates.

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

jeepers3327:
Can I work on this one?

reichert621:
sure @jeepers3327! i'll assign you for now 👍

reichert621:
thanks @jeepers3327! (forgot to close this 😅 )

PR #494, opened 2020-12-29
Sets closed_at field on conversation update

jeepers3327:
### Description
* Add closed_at field to `conversations` table
* Update the changeset of `ChatApi.Conversations.Conversation` module to include `closed_at`
* Instead of setting the value of `closed_at` field after conversation update, I created a function inside `ChatApi.Conversations.Conversation` module to set `closed_at` depending on the `status` passed and include it in changeset pipeline.

### Issue

This handles #486 

## Checklist

- [X] Ran `mix format`


test/chat_api/conversations_test.exs
```diff
@@ -286,6 +286,20 @@ defmodule ChatApi.ConversationsTest do
       refute conv2.archived_at
     end
 
+    test "update_conversation/2 sets the closed_at field based on updated status", %{
```
reichert621, line 289: hooray for tests 🎉 can we add one to `conversation_controller_text.exs` as well?

reichert621 (commented):
looks great @jeepers3327! thanks for the quick turnaround on this 🚀 

just one small request -- would love to have a test in `conversation_controller_test.exs` that verifies this change works as expected too :)

test/chat_api/conversations_test.exs
jeepers3327: Can I ask how should I perform the controller test for these changes? Do I need to add `closed_at` field in "basic.json" in `ChatApiWeb.ConversationView`? I'm not sure about this one.

Hope for your guidance on this one so I could perform the remaining tests.
Thank you.

reichert621: yup you can add `closed_at` to both `basic.json` and `expanded.json`, and then make sure this value is changed correctly when the `status` field is updated

jeepers3327: Would it be ok to just assert the `closed_at` field like  `assert closed_at != nil` if status is "closed" while ` assert closed_at == nil` if status is "open" ?

reichert621: yup that's fine 👍

reichert621 (approved):
looks great, thank you @jeepers3327! 🎉

Issue #487, opened 2020-12-21
Figure out the best way to determine how long it takes to reply to a conversation


reichert621:
**Problem**
Our users want to measure how long it takes from a conversation being created to the first reply from an agent.

At the moment we could determine this by querying conversations and messages and comparing `conversation.inserted_at` with `message.inserted_at` for the first message that comes from an agent (i.e. the first message with a `user_id`).

However I'm a little worried that this might end up being an expensive query, so I'm not sure if it's worth adding a new field (e.g. `first_replied_at`?) to track when this happens.

**Solution**
I'm not sure what the best solution is, but I imagine it's either:
- Write a query that returns `conversation.inserted_at` vs `first_agent_reply_message.inserted_at`
- OR, add a new field to the `conversations` table (e.g. `first_replied_at`) and set this field whenever we detect the initial agent reply to a conversation
  - Then, to query and compare these fields is trivial

**Testing**
We should definitely write some tests to make sure this works as expected, regardless of which approach we take :) 

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

cheeseblubber:
I prefer the solution of adding a first_replied_at field mainly because the queries is going to be pretty expensive as the number of conversations of a company grows and it is probably better to cache the calculation so we don't have to do it every time.

I plan on setting the first_replied_at field when the first message from the agent is first sent on message_controller's create function

cheeseblubber:
@reichert621 Probably plan on creating a function in notification_channel.ex "maybe_update_first_replied_at" that checks if a conversation has a "first_replied_at" field and then update the field there. 

Other options were doing it in the message controller on create but it felt like notification_channel.ex fits a bit better

reichert621:
@cheeseblubber sounds good, tbh there's probably a bit of refactoring we should do eventually to handle the fact that we can create messages from 3 sources as an "agent": the dashboard (via websockets, `notification_channel`), the API (via the `message_controller`), and Slack (via `slack_controller` webhook)

i think starting off in the `notification_channel` makes the most sense 👍 

one nice thing about this update is that we can always backfill data pretty easily (since we can just query all conversations with messages and see when the first message from an agent shows up)

[...]

Issue #473, opened 2020-12-15
As a customer support manager I want to see the response time of my agents


cheeseblubber:
**Problem**
I want to know how long on average it takes to respond on the customer support team level and individual level

PR #495, opened 2020-12-29
Track agent response time

cheeseblubber:
### Description

This is a WIP pull request just wanted to open this for some feedback while I finish up

### Issue

https://github.com/papercups-io/papercups/issues/487

https://github.com/papercups-io/papercups/issues/473

- [x] Ran `mix format`

### DIFF
--- a/assets/src/components/reporting/ReportingDashboard.tsx
+++ b/assets/src/components/reporting/ReportingDashboard.tsx
@@ -7,6 +7,7 @@ import MessagesPerDayChart from './MessagesPerDayChart';
 import MessagesPerUserChart from './MessagesPerUserChart';
 import MessagesSentVsReceivedChart from './MessagesSentVsReceivedChart';
 import MessagesByDayOfWeekChart from './MessagesByDayOfWeekChart';
+import FirstResponseTimeByWeekChart from './FirstResponseTimeByWeekChart';
 import CustomerBreakdownChart from './CustomerBreakdownChart';
 import {ReportingDatum} from './support';
 import logger from '../../logger';
@@ -34,6 +35,11 @@ interface CustomerBreakdownCount {
   count: number;
 }
 
+type ResponseTimeByWeekDay = {
+  day: string;
+  average: number;
+};
+
 type Props = {};
 type State = {
   fromDate: dayjs.Dayjs;
@@ -48,6 +54,7 @@ type State = {
   customerBreakdownByOs: Array<CustomerBreakdownCount>;
   customerBreakdownByTimezone: Array<CustomerBreakdownCount>;
   averageTimeToFirstRespond: number;
+  firstResponseTimeByWeekday: Array<ResponseTimeByWeekDay>;
 };
 
 class ReportingDashboard extends React.Component<Props, State> {
@@ -64,6 +71,7 @@ class ReportingDashboard extends React.Component<Props, State> {
     customerBreakdownByOs: [],
     customerBreakdownByTimezone: [],
     averageTimeToFirstRespond: 0,
+    firstResponseTimeByWeekday: [],
   };
 
   componentDidMount() {
@@ -92,6 +100,7 @@ class ReportingDashboard extends React.Component<Props, State> {
       customerBreakdownByOs: data?.customer_breakdown_by_os || [],
       customerBreakdownByTimezone: data?.customer_breakdown_by_time_zone || [],
       averageTimeToFirstRespond: data?.average_time_to_first_respond || 0,
+      firstResponseTimeByWeekday: data?.first_response_time_by_weekday || [],
     });
   };
 
@@ -174,6 +183,16 @@ class ReportingDashboard extends React.Component<Props, State> {
     });
   };
 
+  formatResponseTimeByWeekDay = (stats: Array<ResponseTimeByWeekDay>) => {
+    return stats.map((ele) => {
+      // use minutes instead of seconds when rendering the chart
+      return {
+        average: Math.round((ele.average / 60) * 100) / 100,
+        day: ele.day.slice(0, 3),
+      };
+    });
+  };
+
   handleDateRangeUpdated = (range: any) => {
     const [fromDate, toDate] = range;
 
@@ -191,7 +210,7 @@ class ReportingDashboard extends React.Component<Props, State> {
     //time would look like 00:01:20 if on average it takes 80 seconds to respond
     const time = new Date(seconds * 1000).toISOString().substr(11, 8);
     const [hour, minute, second] = time.split(':');
-    return `${hour} hr ${minute} min ${second} sec`;
+    return `${hour} h ${minute} m ${second} s`;
   };
 
   render() {
@@ -202,6 +221,7 @@ class ReportingDashboard extends React.Component<Props, State> {
       customerBreakdownByOs = [],
       customerBreakdownByTimezone = [],
       averageTimeToFirstRespond = 0,
+      firstResponseTimeByWeekday = [],
     } = this.state;
     const dailyStats = this.formatDailyStats();
     const userStats = this.formatUserStats();
@@ -221,6 +241,10 @@ class ReportingDashboard extends React.Component<Props, State> {
 
     const responsTimeStats = this.formatResponseTimeStats(
       averageTimeToFirstRespond
+    );
+
+    const responseTimeByWeekDay = this.formatResponseTimeByWeekDay(
+      firstResponseTimeByWeekday
     );
 
     return (
@@ -253,12 +277,15 @@ class ReportingDashboard extends React.Component<Props, State> {
               </Box>
               <Box>
                 <Box mb={2}>
-                  <Text>Average 1st Response </Text>
+                  <Text>average first response time: </Text>
+                  <Text style={{fontWeight: 'bold'}}>{responsTimeStats}</Text>
                 </Box>
-                <Text>{responsTimeStats}</Text>
               </Box>
+              <FirstResponseTimeByWeekChart data={responseTimeByWeekDay} />
             </Box>
-
+          </Flex>
+          <Divider />
+          <Flex mx={-3} mb={4}>
             <Box mb={4} mx={3} sx={{height: 320, maxWidth: '50%', flex: 1}}>
               <Box mb={2}>
                 <Text strong>New messages per day</Text>
--- a/assets/src/components/reporting/ReportingDashboard.tsx
+++ b/assets/src/components/reporting/ReportingDashboard.tsx
@@ -47,6 +47,7 @@ type State = {
   customerBreakdownByBrowser: Array<CustomerBreakdownCount>;
   customerBreakdownByOs: Array<CustomerBreakdownCount>;
   customerBreakdownByTimezone: Array<CustomerBreakdownCount>;
+  averageTimeToFirstRespond: number;
 };
 
 class ReportingDashboard extends React.Component<Props, State> {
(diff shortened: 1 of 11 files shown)


cheeseblubber:
Screen Shot 2020-12-29 at 5 25 54 PM

lib/chat_api_web/controllers/reporting_controller.ex
```diff
@@ -21,11 +21,13 @@ defmodule ChatApiWeb.ReportingController do
         conversations_by_date: Reporting.count_conversations_by_date(account_id, filters),
         messages_per_user: Reporting.count_messages_per_user(account_id, filters),
         messages_by_weekday: Reporting.count_messages_by_weekday(account_id, filters),
+        first_response_time_by_weekday: Reporting.first_response_time_by_weekday(account_id),
```
reichert621, line 24: let's make sure to pass in `filters` here (which includes a from_date/to_date)

lib/chat_api_web/controllers/reporting_controller.ex
```diff
@@ -21,11 +21,13 @@ defmodule ChatApiWeb.ReportingController do
...
         sent_messages_by_date: Reporting.count_sent_messages_by_date(account_id, filters),
         received_messages_by_date: Reporting.count_received_messages_by_date(account_id, filters),
         customer_breakdown_by_browser:
           Reporting.get_customer_breakdown(account_id, :browser, filters),
         customer_breakdown_by_os: Reporting.get_customer_breakdown(account_id, :os, filters),
+        average_time_to_first_respond: Reporting.average_first_replied_time(account_id),
```
reichert621, line 30: (same here)

test/chat_api/reporting_test.exs
```diff
@@ -317,6 +335,83 @@ defmodule ChatApi.ReportingTest do
...
+
+      # Monday
+      insert(:conversation,
+        account: account,
+        inserted_at: ~N[2020-10-05 10:00:00],
+        first_replied_at: ~N[2020-10-05 11:00:03]
+      )
+
+      # Tuesday
+      insert_pair(:conversation,
```
reichert621, line 362: just curious, why `insert_pair` for this one?

lib/chat_api/reporting.ex
```diff
@@ -40,6 +40,26 @@ defmodule ChatApi.Reporting do
...
+  end
+
+  def compute_average_replied_time(conversations) do
+    conversations
+    |> Enum.map(fn conv -> Time.diff(conv.first_replied_at, conv.inserted_at) end)
+    |> average()
+  end
+
+  def average(list) do
+    total = Enum.reduce(list, 0, fn x, acc -> x + acc end)
```
reichert621, line 59: nit -- we can just use `Enum.sum(list)` here

reichert621 (commented):
@cheeseblubber looks good! 🚀  📈 

just some high-level feedback:
- let's keep this PR focused on `first_replied_at` since https://github.com/papercups-io/papercups/pull/494 is already handling `closed_at`
- slight preference for keeping this PR just focused on the backend + reporting APIs + tests, without worrying about the UI yet, to keep the review a little easier
- rather than getting average response time by day of week, i feel like it would probably be more useful for our users to get average response time per agent, or on a weekly/monthly basis... what do you think?

(I also think some warnings are popping up because you might need to rebase on master 🤔)

lib/chat_api/reporting.ex
```diff
@@ -40,6 +40,26 @@ defmodule ChatApi.Reporting do
     |> Repo.all()
   end
 
+  def average_first_replied_time(account_id, filters \\ %{}) do
```
reichert621, line 43: let's add typespecs to these functions, it should help with warnings/readability/IDE experience improvements 👍 

e.g.
```elixir
@spec average_first_replied_time(binary(), map()) :: float()
```

also I wonder if we should make it clearer that the output of this is in `seconds` (vs `ms` or something else)... maybe we could name it something like `average_seconds_to_first_reply` or `average_time_to_first_reply_in_seconds`? not sure 🤷

lib/chat_api/reporting.ex
```diff
@@ -40,6 +40,26 @@ defmodule ChatApi.Reporting do
...
+  end
+
+  def compute_average_replied_time(conversations) do
+    conversations
+    |> Enum.map(fn conv -> Time.diff(conv.first_replied_at, conv.inserted_at) end)
+    |> average()
+  end
+
+  def average(list) do
+    Enum.sum(list) / max(length(list), 1)
```
reichert621, line 60: not important but for the sake of learning elixir -- one common technique people use in elixir is pattern matching to handle special cases (e.g. an empty list, which in this case would result in a denominator of zero)

for example, in this case we could write this like
```elixir
@spec average([integer()]) :: float()
def average([]), do: 0

def average(list) do
  Enum.sum(list) / length(list)
end
```

(just a useful pattern to be aware of, since elixir doesn't support early returns)

lib/chat_api/reporting.ex
```diff
@@ -101,13 +121,38 @@ defmodule ChatApi.Reporting do
...
   end
 
+  # TODO some duplication here with group by date might be good to refactor
+  defp average_grouped_by_date(query, field \\ :inserted_at) do
+    query
+    |> group_by([r], fragment("date(?)", field(r, ^field)))
+    |> select([r], %{
+      date: fragment("date(?)", field(r, ^field)),
+      count: count(r.id),
+      # avg doesn't do anything but raises a grouping_error when I remove it...
```
reichert621, line 150: hmm that's a little concerning... would be nice to figure out the underlying issue here?

test/chat_api/reporting_test.exs
```diff
@@ -91,6 +91,24 @@ defmodule ChatApi.ReportingTest do
                Reporting.count_conversations_by_date(account.id)
     end
 
+    test "average_first_replied_time gets the average seconds it takes to respond", %{
```
reichert621, line 94: would be good to also test cases with more variability in reply time, as well as cases where some conversations have `first_replied_at: nil`

test/chat_api/reporting_test.exs
```diff
@@ -317,6 +335,83 @@ defmodule ChatApi.ReportingTest do
...
+      )
+
+      assert [
+               %{day: "Monday", average: 3663.0},
+               %{day: "Tuesday", average: 3723.0},
+               %{day: "Wednesday", average: 123.0},
+               %{day: "Thursday", average: 123.0},
+               %{day: "Friday", average: 3.0},
+               %{day: "Saturday", average: 3.0},
+               %{day: "Sunday", average: 3603.0}
```
reichert621, line 410: i wonder if we should include a `unit: "seconds"` field or something to make it clearer that the `average` is in seconds?

test/chat_api_web/channels/notification_channel_test.exs
```diff
@@ -42,4 +44,44 @@ defmodule ChatApiWeb.NotificationChannelTest do
     broadcast_from!(socket, "broadcast", %{"some" => "data"})
     assert_push "broadcast", %{"some" => "data"}
   end
+
+  test "conversation first replied at is updated", %{
```
reichert621, line 48: nice! we should probably also test that this only happens for the first reply message

lib/chat_api/reporting.ex
cheeseblubber: Sounds good thanks!

cheeseblubber:
Screen Shot 2021-01-01 at 7 43 50 PM
I m gonna keep the client side in here for now since there isn't too much code to review and made a separate issue for tracking response time by agent https://github.com/papercups-io/papercups/issues/499

assets/src/components/reporting/ReportingDashboard.tsx
```diff
@@ -207,6 +239,14 @@ class ReportingDashboard extends React.Component<Props, State> {
       'time_zone'
     );
 
+    const responsTimeStats = this.formatResponseTimeStats(
```
reichert621, line 242: nit: typo (`responsTimeStats` -> `responseTimeStats`)

assets/src/components/reporting/ReportingDashboard.tsx
```diff
@@ -183,13 +205,23 @@ class ReportingDashboard extends React.Component<Props, State> {
     );
   };
 
+  formatResponseTimeStats = (responseTime: number) => {
+    const seconds = Math.round(responseTime);
```
reichert621, line 209: nit: is this necessary?

assets/src/components/reporting/ReportingDashboard.tsx
```diff
@@ -183,13 +205,23 @@ class ReportingDashboard extends React.Component<Props, State> {
...
+    //time would look like 00:01:20 if on average it takes 80 seconds to respond
+    const time = new Date(seconds * 1000).toISOString().substr(11, 8);
```
reichert621, line 211: i think `substr` is a legacy function (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#The_difference_between_substring_and_substr)

but we can just do
```js
new Date(seconds * 1000).toISOString().substring(11, 19);
```

also since this is a pure function that doesn't require any component state and seems like it could be useful in other cases outside of the context of this component, we should probably just move it out into `/reporting/support.ts` so we can test it more easily (once we have more frontend tests 😛) -- what do you think?

assets/src/components/reporting/ReportingDashboard.tsx
```diff
@@ -171,6 +183,16 @@ class ReportingDashboard extends React.Component<Props, State> {
     });
   };
 
+  formatResponseTimeByWeekDay = (stats: Array<ResponseTimeByWeekDay>) => {
+    return stats.map((ele) => {
```
reichert621, line 187: not important, but one useful js/ts thing you can do here is just destructure `(ele) =>` into `({average, day}) =>` (might improve readability a bit, since `ele` isn't particularly descriptive)

reichert621 (changes requested):
looks good! just some minor stuff and then should be good to go 👍

assets/src/components/reporting/ReportingDashboard.tsx
cheeseblubber: This is slightly more accurate than if we didn't have it. If it is 25.8 seconds without it it would become 25 seconds instead of 26 seconds. Its pretty simple piece of code and makes it more accurate so I think we should just keep it in

assets/src/components/reporting/ReportingDashboard.tsx
cheeseblubber: Yeah sounds good I'll move it out

assets/src/components/reporting/ReportingDashboard.tsx
```diff
@@ -171,6 +183,16 @@ class ReportingDashboard extends React.Component<Props, State> {
     });
   };
 
+  formatResponseTimeByWeekDay = (stats: Array<ResponseTimeByWeekDay>) => {
+    return stats.map((average: string, day) => {
```
reichert621, line 187: @cheeseblubber this breaks the frontend build -- should be `({average, day})` in curly brackets

(if you're using vscode, it might be helpful to install the `TSLint` plugin and the `Prettier - Code formatter` plugin to help catch things like this)

reichert621 (commented):
@cheeseblubber one comment about some code that breaks the frontend build -- we should probably add better guardrails to prevent this, but in the meantime the `TSLint` plugin in vscode should be helpful

Issue #515, opened 2021-01-11
The public `/api/customers/identify` endpoint should factor in the current `host`


reichert621:
**Problem**
For customers who include Papercups in their self-hosted open-source projects, there is potential for their user IDs to conflict with other deployments of the project (e.g. if the `users` table has auto-incrementing integer IDs).

In the meantime, to work around this customers can fix this by updating the `external_id` set on the widget to be unique across environments (e.g. by including the hostname in the ID, or some other unique identifier)

But we should obvious fix it on our end too!

**Solution**
I think the easiest solution for now is to use the `host` information on the HTTP request as an additional filter when searching for a matching customer here: https://github.com/papercups-io/papercups/blob/master/lib/chat_api_web/controllers/customer_controller.ex#L47

One potential issue with matching against the host name EXACTLY is that it would prevent user conversations from being shared across subdomains, but we can always fix that later. I'd rather err on the side of being _more_ restrictive than less.

reichert621:
after thinking about this a bit more, it might be best to avoid using explicit customer fields (like id and email) to figure out which conversations to load in general, and just stick with the more opaque token we cache in the browser

this has the downside of not loading conversation history when a user switches browsers/devices, but it might be a reasonable trade-off

paolodamico:
As a suggestion you could also **optionally** delegate the token logic to the client (e.g. PostHog). Then on the client side we can store a token (on our DB) with enough randomness to be secure enough. An even better approach would be the ability to provide a user ID with a tamper-proof signature that you can validate async (e.g. a JWT). With this, the client would have the option of supporting end users moving between devices/browsers without losing their chat history without compromising security. Though I would do this as an improvement after fixing the underlying bug.

reichert621:
@paolodamico i like that idea :) but i'm still not completely sure i understand how that would work in practice -- would papercups be issuing the JWT, or would you guys?

paolodamico:
I've been giving it more thought and doing the shared secret or some sort of async signing might be an overkill. Particularly, if we consider for instance a JWT signed with `HS256`, the signature hash length would be 32 bytes. From this you could simply **allow the client the option to provide the unique ID** and maybe just clarify that the ID should have enough entropy (be generated by a cryptographically secure random generator and I would assume the 32 byte length), and that overriding this has security implications. You could consider enforcing on your end a certain length to make sure it's used correctly, but this is not enough to guarantee entropy and might not be worth the trouble.

[...]

PR #516, opened 2021-01-11
Quick fix for securing `/api/customers/identify` better

reichert621:
### Description

Quick fix for securing `/api/customers/identify` better by including filters for `email` and `host` in the query, to prevent malicious users from trying to guess `external_id` + account token combos.

(Note that we'll want to come up with a better long term solution here, but this should prevent the collisions we're seeing at the moment.)

### Issue

https://github.com/papercups-io/papercups/issues/515

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api/customers.ex
+++ b/lib/chat_api/customers.ex
@@ -41,18 +41,21 @@ defmodule ChatApi.Customers do
     Customer |> Repo.get!(id) |> Repo.preload(:tags)
   end
 
-  @spec find_by_external_id(binary(), binary()) :: Customer.t() | nil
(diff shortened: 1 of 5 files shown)


aguilaair:
Hey @reichert621 dies this require any changes or from the chat widget side?

reichert621:
@aguilaair it should be backward compatible and won't break anything on your side, but I would recommend doing a similar update as in https://github.com/papercups-io/chat-window/pull/25/files 

 over the next week or two we may introduce some breaking changes for added security, but I'll make sure to tag you so you're alerted 👍

PR #522, opened 2021-01-13
Make it possible to customize username and photo in Papercups Slack app

reichert621:
### Description

Adds the scope to make it possible to customize username and photo in Papercups Slack app

### Issue

Users want to show up as "themselves", rather than a bot/app

### Screenshots

Screen Shot 2021-01-13 at 12 43 54 PM

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api/slack/notifications.ex
+++ b/lib/chat_api/slack/notifications.ex
@@ -197,16 +197,16 @@ defmodule ChatApi.Slack.Notifications do
     end
   end
 
+  @papercups_app_name "Papercups"
+  @papercups_icon_url "https://s3-us-west-2.amazonaws.com/slack-files2/avatars/2021-01-05/1626939067681_3e27968eb3657d7167e5_132.png"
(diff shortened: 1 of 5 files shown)


lib/chat_api/slack/notifications.ex
```diff
@@ -143,7 +143,9 @@ defmodule ChatApi.Slack.Notifications do
       message = %{
         "text" => format_slack_message_text(message),
         "channel" => thread.slack_channel,
-        "thread_ts" => thread.slack_thread_ts
+        "thread_ts" => thread.slack_thread_ts,
+        "username" => format_user_name(user),
```
reichert621, line 147: (still need to test how this is handled when the required scopes are missing... don't merge until that's verified)

reichert621: tested, seems to just ignore these fields when scope is missing 👍

PR #544, opened 2021-01-23
fix padding height on all conversations css

cheeseblubber:
Before:
Screen Shot 2021-01-23 at 9 01 41 AM

After:
Screen Shot 2021-01-23 at 9 01 23 AM

### DIFF
--- a/assets/src/components/conversations/ConversationsContainer.tsx
+++ b/assets/src/components/conversations/ConversationsContainer.tsx
@@ -341,7 +341,12 @@ class ConversationsContainer extends React.Component<Props, State> {
             left: 220,
           }}
         >
-          <Box p={3} sx={{borderBottom: '1px solid #f0f0f0'}}>
+          <Box
+            pl={3}
+            pt={3}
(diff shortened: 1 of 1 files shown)


cheeseblubber:
The two rows not being aligned has bothered me for a while.

reichert621:
yeah it's slightly annoying that it's not aligned (especially since it gets so close), but the conversation header is sized dynamically in order to keep the text vertically centered, so if both name and email are present it will look like this (slightly "taller" rather than shorter):
Screen Shot 2021-01-26 at 5 24 15 PM

or on smaller screens, like this:
Screen Shot 2021-01-26 at 5 28 45 PM

i imagine in the near future we'll be adding more information/functionality to these components anyway, so i don't think getting the alignment perfect is worth the headache... here are Intercom's headers for reference (notice none of them are aligned):
Screen Shot 2021-01-26 at 5 26 44 PM

closing this for now, but feel free to reopen if you have any other ideas for how to improve the design here!

PR #556, opened 2021-02-01
Load previous messages in Slack thread when replying to bot message

reichert621:
### Description

This PR is a first step toward loading the whole thread when someone replies to a bot message. (This feature is not widely used yet, so we'll be releasing improvements to this incrementally.)

## Checklist

- [ ] Ran `mix format`


lib/chat_api/slack/event.ex
```diff
@@ -0,0 +1,432 @@
+defmodule ChatApi.Slack.Event do
```
reichert621, line 1: (just setting up new module to hold all our slack event handlers... this code is basically all copied from the slack controller)

PR #613, opened 2021-03-03
Make it possible to delete a user and all of its data via API 

### LINKED ISSUE #600: Make it possible to delete a user and all of its data via API
**Problem**
We want to be able to delete a user record and all of its associated data. Eventually we'll add a UI to enable this, but to start off we should be able to do it safely and programmatically.

(This will probably require some database migrations that improve how associated records are deleted... at the moment some resources are not automatically deleted when a parent resource is attempted to be deleted.)

This will also be necessary for GDPR compliance.

**Solution**
Set up an API for `DELETE /api/users/:id` (or similar) which deletes a user and all of its data.

**Testing**
Write automated tests and verify that all the user data is deleted, and no other user is affected.

fmterrorf:
### Description

- [x] DB migration to fix how associated resources are deleted when `user` is deleted.

### Issue

Closes #600 

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/priv/repo/migrations/20210302011754_add_cascade_deletes_to_user_references.exs
+++ b/priv/repo/migrations/20210302011754_add_cascade_deletes_to_user_references.exs
@@ -5,6 +5,7 @@ defmodule ChatApi.Repo.Migrations.AddCascadeDeletesToUserReferences do
       drop constraint("messages", "messages_user_id_fkey")
       drop constraint("conversations", "conversations_assignee_id_fkey")
       drop constraint("user_profiles", "user_profiles_user_id_fkey")
+      drop constraint("user_settings", "user_settings_user_id_fkey")
       drop constraint("google_authorizations", "google_authorizations_user_id_fkey")
       drop constraint("tags", "tags_creator_id_fkey")
       drop constraint("conversation_tags", "conversation_tags_creator_id_fkey")
@@ -16,10 +17,14 @@ defmodule ChatApi.Repo.Migrations.AddCascadeDeletesToUserReferences do
       end
 
       alter table(:conversations) do
-        modify(:assignee_id, references(:users, on_delete: :delete_all))
+        modify(:assignee_id, references(:users, on_delete: :nilify_all))
       end
 
       alter table(:user_profiles) do
+        modify(:user_id, references(:users, on_delete: :delete_all))
+      end
+
+      alter table(:user_settings) do
         modify(:user_id, references(:users, on_delete: :delete_all))
       end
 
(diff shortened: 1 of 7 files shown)


priv/repo/migrations/20210302011754_add_cascade_deletes_to_user_references.exs
```diff
@@ -0,0 +1,99 @@
...
+    alter table(:google_authorizations) do
+      modify(:user_id, references(:users, on_delete: :delete_all))
+    end
+
+    alter table(:tags) do
+      modify(:creator_id, references(:users, on_delete: :nilify_all))
+    end
+
+    alter table(:conversation_tags) do
+      modify(:creator_id, references(:users, on_delete: :delete_all))
```
reichert621, line 40: i think this one can just be `nilify_all` as well :+1:

reichert621 (commented):
great work @fmterrorf! 🎉 

just one small thing and then it should be good to go :+1:

a8t:
@fmterrorf I'm a novice with Elixir. Reading this PR helped a lot. May I ask you some simple questions about this PR to further help me learn?

fmterrorf:
@a8t yeah for sure! I'd be happy answer any questions

reichert621 (approved):
great work @fmterrorf! thanks for the help on this 🎉 🚀

a8t (commented):
Was this migration generated automatically somehow, or did you write it by hand?

test/chat_api/users_test.exs
```diff
@@ -1,12 +1,45 @@
...
   import ChatApi.Factory
 
+  alias ChatApi.Repo
   alias ChatApi.Users
   alias ChatApi.Users.{User, UserProfile, UserSettings}
 
+
+  describe "user" do
+    setup do
+      {:ok, user: insert(:user)}
```
a8t, line 14: Where does the `insert` function come from?

a8t:
I left 2 comments. Thanks Daven!

test/chat_api/users_test.exs
fmterrorf: The insert function is coming from this https://github.com/papercups-io/papercups/blob/master/test/support/factory.ex 

In `user_test.exs` we have call `  import ChatApi.Factory`.  The `factory.ex` file calls uses [ExMachina.ecto](https://hexdocs.pm/ex_machina/ExMachina.Ecto.html) which gives you that `insert` function

fmterrorf:
Was this migration generated automatically somehow, or did you write it by hand?

I wrote it by hand by creating the migration using [mix ecto.gen.migration (My migration name)](https://hexdocs.pm/ecto_sql/Mix.Tasks.Ecto.Gen.Migration.html), and adding the up/down migration

a8t:
Thanks @fmterrorf. It was useful reference for my recent PR.

PR #617, opened 2021-03-04
Use correct changeset typespec in User context

fmterrorf:
### Description

Forgot to add this on my previous pr here https://github.com/papercups-io/papercups/pull/613, but `Ecto.Changeset.User` should be `Ecto.Changeset.t()`

- [x] Ran `mix format`


reichert621 (approved):
thanks!

Issue #451, opened 2020-12-07
Implement pagination for the `ChatApi.Conversations.list_conversations_by_account/2` method


reichert621:
**Problem**
We currently fetch ALL conversation records when we hit the `ChatApi.Conversations.list_conversations_by_account/2` method (i.e. we set no `limit` or `offset` in our query, nor do we have cursors or anything like that)

This will not scale very well as more and more data gets added.

**Solution**
To start off, let's do the following:
- Implement another version of `ChatApi.Conversations.list_conversations_by_account(account_id, filters)` that paginates the response, ideally using cursors
  - Here's where the method lives in the code: https://github.com/papercups-io/papercups/blob/master/lib/chat_api/conversations.ex#L24
  - Possible library for cursor approach: https://github.com/duffelhq/paginator
- (Note: I have a very minimal understanding of how to use cursors, so I may be wrong that this is the best approach... feel free to push back!)
- This new method should just be called something like `ChatApi.Conversations.list_conversations_by_account(account_id, filters, pagination_options)`, and should only return, let's say, 50 records at a time
- Conversations should be sorted by **most recent message** -- not sure how tricky this will be given it may require some `join` queries on the `messages` table, but maybe we'll just have to add a new field for `last_message_sent_at` or something 🤔 

**Testing**
For the sake of this ticket, DO NOT use this new method in the `GET /api/conversations` endpoint. We just want to set up the method itself, and write some tests showcasing how it works.

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

psolvy:
Hi
I'll take this one)

reichert621:
sounds good @psolvy! I'll assign you for now :)

flmel:
Hi  @psolvy, @reichert621

If we want to use prams such as limit offset  eg. _`%{page_size: 5, page_number: 10}`_ we cant use cursor based pagination.
I have opened a [pr](#490) which uses [scrivener](https://github.com/drewolson/scrivener_ecto) (offset based pagination). 

also this is related with #332

reichert621:
[...]

PR #631, opened 2021-03-12
Paginate list conversation

fmterrorf:
### Description
* Added `Conversation.list_conversations_by_account_paginated/2` function. 

### Issue
related #451 

<details>
  <summary>Generated SQL</summary>
  
  ```sql
SELECT 
  c0."id", 
  c0."status", 
  c0."priority", 
  c0."source", 
  c0."read", 
  c0."archived_at", 
  c0."first_replied_at", 
  c0."closed_at", 
  c0."metadata", 
  c0."assignee_id", 
  c0."account_id", 
  c0."customer_id", 
  c0."inserted_at", 
  c0."updated_at" 
FROM 
  "conversations" AS c0 
  LEFT OUTER JOIN LATERAL (
    SELECT 
      inserted_at 
    FROM 
      messages 
    WHERE 
      conversation_id = c0."id" 
    ORDER BY 
      inserted_at DESC 
    LIMIT 
      1
  ) AS f1 ON TRUE 
WHERE 
  (c0."account_id" = $1) 
  AND (c0."archived_at" IS NULL) 
  AND (
    (
      (f1."inserted_at" = $2) 
      AND (
        (c0."id" < $3) 
        AND $4
      )
    ) 
    OR (
      (
        (f1."inserted_at" < $5) 
        AND $6
      ) 
      AND $7
    )
  ) 
ORDER BY 
  f1 DESC, 
  c0."id" DESC 
LIMIT 
  $8
  ```

</details>

## Checklist

- [x] Ran `mix format`


lib/chat_api/conversations.ex
```diff
@@ -23,6 +23,26 @@ defmodule ChatApi.Conversations do
...
+    |> where(^filter_where(filters))
+    |> where([c], is_nil(c.archived_at))
+    |> join(
+      :left_lateral,
+      [c],
+      f in fragment(
+        "SELECT inserted_at FROM messages WHERE conversation_id = ? ORDER BY inserted_at DESC LIMIT 1",
+        c.id
+      ),
+      as: :messages
```
fmterrorf, line 39: I need to add this binding because `Paginator.paginate` expects that the fields in the `cursor_field` to be in the query aliases

lib/chat_api/conversations.ex
```diff
@@ -483,4 +503,31 @@ defmodule ChatApi.Conversations do
...
+      args,
+      [
+        # These are paginator options from https://hexdocs.pm/paginator/Paginator.html#c:paginate/3-options
+        cursor_fields: [{:messages, :inserted_at}, :id],
+        sort_direction: :desc,
+        fetch_cursor_value_fun: fn
+          schema, {:messages, :inserted_at} ->
+            # Get the `inserted_at` field of the most recent message
+            Enum.sort_by(schema.messages, & &1.inserted_at, :desc)
+            |> Enum.map(& &1.inserted_at)
```
fmterrorf, line 519: I have to take the most recent `message.inserted_at` for each conversation, so that i can use it for the cursor_field

lib/chat_api/conversations.ex
```diff
@@ -23,6 +23,26 @@ defmodule ChatApi.Conversations do
     |> Repo.all()
   end
 
+  @spec list_conversations_by_account_paginated(binary(), map()) :: Paginator.Page.t()
+  def list_conversations_by_account_paginated(account_id, filters \\ %{}) do
```
fmterrorf, line 27: I just realized that you had this comment in the issue 
> This new method should just be called something like ChatApi.Conversations.list_conversations_by_account(account_id, filters, pagination_options), and should only return, let's say, 50 records at a time

Should i change this method to that signature?

lib/chat_api/repo.ex
```diff
@@ -1,7 +1,17 @@
...
+  ]
+
   use Ecto.Repo,
     otp_app: :chat_api,
     adapter: Ecto.Adapters.Postgres
 
   use Scrivener, page_size: 50
+
+  @spec paginate_with_cursor(any, keyword, any) :: Paginator.Page.t()
+  def paginate_with_cursor(queryable, opts \\ [], repo_opts \\ []) do
```
fmterrorf, line 13: https://hexdocs.pm/paginator/Paginator.html#module-use-without-macros
The `Scrivener` macro uses `Repo.paginate`, so i have to use `Paginator` without a macro, and use a different function name

reichert621:
@fmterrorf looks great! i need to take a little more time to review but wanted to get your thoughts on something... after working on this, do you think it would make our lives a lot easier to just add a new field to `conversations` that holds the `last_message_sent_at`/`last_activity_at`, and make sure to update that whenever a message is created?

fmterrorf:
@reichert621 yeah I believe so. I think that will make ordering better

fmterrorf:
Som maybe we could do
1. Add the `last_activity_at` field. Then we would use latest `messages.inserted_at` value for the `conversations.last_activity_at` for existing conversations. This will be a migration
2.  Then implement the pagination

reichert621:
Yup that sounds very reasonable to me

reichert621:
feel free to handle those two steps you listed in separate PRs if that's easier btw!

fmterrorf:
Going to add a separate PR for implementing last_activity and pagination

Issue #623, opened 2021-03-10
Add boilerplate CRUD APIs for `canned_responses`


reichert621:
**Problem**
Users have been asking for canned responses, and the schema is pretty straightforward, so let's start setting it up!

**Solution**
Set up the db migrations and boilerplate CRUD (create/read/update/delete) methods for a `canned_responses` table.

For an example of how to do this, you can probably use https://hexdocs.pm/phoenix/Mix.Tasks.Phx.Gen.Json.html (or see https://github.com/papercups-io/papercups/blob/master/docs/examples/add_conversations_api.md for some additional examples of this in action)

The schema should look something like this:
- `name`: required/non-nullable string field that represents the "code" for the canned response
  - Note that this should be unique per account (i.e. there should not be two `canned_response` records with the same `name` within a single `account`
- `content`: required/non-nullable text field that represents the full message of the response
- `account_id`: required/non-nullable foreign key for `accounts`

An example record might look like:
- `name: "pricing"`
- `content: "We have a free tier for X users! For additional pricing options, check out papercups.io/pricing"`
- `account_id: "__SOME_ACCOUNT_UUID__"`

**Additional context**
We can worry about the UI/UX for this later.

**Testing**
Please include automated tests! (A bunch of these should be autogenerated)

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

a8t:
What if the user wants variables (or whatever we want to call it) for interpolating into the message? Like `content: "Hi %user_name%, nice to meet you."` - any thoughts on this? Would probably be a next-iteration feature.

reichert621:
seems like a good idea! but i don't think it needs to be part of this ticket :+1:

jeepers3327:
Hi, is this still unassigned?.
I'd like to work on this one later after my work.

a8t:
@jeepers3327 I'm starting work on it now, but I don't mind if we both do and your PR is the one that goes in.

reichert621:
@jeepers3327 @a8t I'll assign you both for now :+1: happy to review both PRs and provide feedback 🤓

PR #632, opened 2021-03-12
Add CRUD API for canned_responses

a8t:
### Description

Adds migration, tests, schema, context, controllers, and views for a new `canned_responses` CRUD API.

Includes tests and code for preventing creation and updating of `canned_responses` that would have a duplicate `:name` per account.

### Related issue

#623 

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api_web/controllers/canned_response_controller.ex
+++ b/lib/chat_api_web/controllers/canned_response_controller.ex
@@ -1,0 +1,45 @@
+defmodule ChatApiWeb.CannedResponseController do
+  use ChatApiWeb, :controller
+
+  alias ChatApi.CannedResponses
+  alias ChatApi.CannedResponses.CannedResponse
+
+  action_fallback ChatApiWeb.FallbackController
+
+  def index(conn, _params) do
+    canned_responses = CannedResponses.list_canned_responses()
+    render(conn, "index.json", canned_responses: canned_responses)
+  end
+
+  def create(conn, %{"canned_response" => canned_response_params}) do
+    with {:ok, %CannedResponse{} = canned_response} <-
+           CannedResponses.create_canned_response(canned_response_params) do
+      conn
+      |> put_status(:created)
+      |> put_resp_header("location", Routes.canned_response_path(conn, :show, canned_response))
+      |> render("show.json", canned_response: canned_response)
+    end
+  end
+
+  def show(conn, %{"id" => id}) do
+    canned_response = CannedResponses.get_canned_response!(id)
+    render(conn, "show.json", canned_response: canned_response)
+  end
+
+  def update(conn, %{"id" => id, "canned_response" => canned_response_params}) do
+    canned_response = CannedResponses.get_canned_response!(id)
+
+    with {:ok, %CannedResponse{} = canned_response} <-
+           CannedResponses.update_canned_response(canned_response, canned_response_params) do
+      render(conn, "show.json", canned_response: canned_response)
+    end
+  end
+
+  def delete(conn, %{"id" => id}) do
+    canned_response = CannedResponses.get_canned_response!(id)
+
+    with {:ok, %CannedResponse{}} <- CannedResponses.delete_canned_response(canned_response) do
+      send_resp(conn, :no_content, "")
+    end
+  end
+end
--- a/lib/chat_api_web/controllers/canned_response_controller.ex
+++ b/lib/chat_api_web/controllers/canned_response_controller.ex
@@ -39,14 +39,21 @@ defmodule ChatApiWeb.CannedResponseController do
   @spec show(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def show(conn, %{"id" => id}) do
     canned_response = CannedResponses.get_canned_response!(id)
-    render(conn, "show.json", canned_response: canned_response)
+
+    with :ok <- is_authorized?(conn.assigns.current_user, canned_response) do
+      render(conn, "show.json", canned_response: canned_response)
+    end
   end
 
   def update(conn, %{"id" => id, "content" => content, "name" => name}) do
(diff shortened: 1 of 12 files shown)


a8t:
Looking through other controllers, I think I need to use `conn.assigns.current_user` to ensure that other users can't randomly edit anyone else's `canned_responses`.  This probably also lets me remove `account_id` from the interface.

lib/chat_api_web/controllers/canned_response_controller.ex
```diff
@@ -0,0 +1,45 @@
+defmodule ChatApiWeb.CannedResponseController do
+  use ChatApiWeb, :controller
+
+  alias ChatApi.CannedResponses
+  alias ChatApi.CannedResponses.CannedResponse
+
+  action_fallback ChatApiWeb.FallbackController
+
+  def index(conn, _params) do
```
reichert621, line 9: @a8t great work on all this! 

main feedback/thoughts:
- for all of these controller methods, make sure we use the `conn.assigns.current_user`'s information to filter the records we're querying
- for example, this `index` method as it is right now would fetch ALL records, rather than just the ones for the authenticated account
- also, for `create` and `update` methods, we need to make sure that the `account_id` always matches the currently authenticated user's account -- we can do this by doing a `Map.merge(params, %{"account_id" => conn.assigns.user.account_id})`, or raising a 403/forbidden error if someone attempts to pass in a different `account_id`

and if you don't mind, it would be great to add typespecs to these methods as well!

(let me know if any of that doesn't make sense or isn't clear)

reichert621 (changes requested):

lib/chat_api_web/controllers/canned_response_controller.ex
a8t: Oops, I didn't see your comment before posting my Slack message earlier, lol.

a8t: I waffled about the best way to do this for a bit, but I updated the PR. please check the `is_authorized?` function.

lib/chat_api_web/controllers/canned_response_controller.ex
```diff
@@ -0,0 +1,84 @@
...
+
+    with :ok <- is_authorized?(conn.assigns.current_user, canned_response),
+         {:ok, %CannedResponse{}} <- CannedResponses.delete_canned_response(canned_response) do
+      send_resp(conn, :no_content, "")
+    end
+  end
+
+  defp is_authorized?(user, canned_response) do
+    cond do
+      user.role == "admin" ->
```
reichert621, line 74: sorry the docs on user roles aren't great 😅 but we don't want admin users to have this much access -- admins just exist within accounts (not the whole platform) and have a bit more capabilities that regular users, but in this case the permissions should be the same (i.e. both "users" and "admin" can access the same canned response resources)

lib/chat_api_web/router.ex
```diff
@@ -114,6 +114,7 @@ defmodule ChatApiWeb.Router do
     resources("/tags", TagController, except: [:new, :edit])
     resources("/browser_sessions", BrowserSessionController, except: [:create, :new, :edit])
     resources("/personal_api_keys", PersonalApiKeyController, except: [:new, :edit, :update])
+    resources("/canned_responses", CannedResponseController)
```
reichert621, line 117: nit - we should exclude `:new` and `:edit` here:
```elixir
resources("/canned_responses", CannedResponseController, except: [:new, :edit])
```

("new" and "edit" are primarily used for full-stack phoenix applications, where we render those pages on the server... for a JSON API they aren't necessary :+1:)

test/chat_api_web/controllers/tag_controller_test.exs
```diff
@@ -66,6 +66,23 @@ defmodule ChatApiWeb.TagControllerTest do
       conn = put(authed_conn, Routes.tag_path(authed_conn, :update, tag), tag: @invalid_attrs)
       assert json_response(conn, 422)["errors"] != %{}
     end
+
+    # test "renders errors when editing another account's tag",
```
reichert621, line 70: should we remove this?

reichert621 (changes requested):
@a8t a few more small comments! i like the https://github.com/schrockwell/bodyguard pattern :+1:

i'm gonna merge https://github.com/papercups-io/papercups/pull/633 for now to give others an opportunity to contribute, but that PR is currently missing the `authorize` fixes you implemented here, so if you want to set up a followup PR fixing those security issues that would be amazing!

lib/chat_api_web/controllers/canned_response_controller.ex
reichert621: btw some super tiny nitpicks:
- since elixir supports `?` in method names, prefixing with `is_` is usually unnecessary (though i do it often too, coming from JS/TS)
- it's generally expected that methods that end in `?` return a boolean value (rather than `:ok` vs `:error`)
- assuming we wanted to keep the logic in this method the same, it would be a little bit more idiomatic as a `case` statement:

```elixir
def authorize(user, %CannedResponse{account_id: account_id}) do
  case user do
    %{role: "admin"} -> :ok
    %{account_id: ^account_id} -> :ok
    _ -> {:error, :forbidden}
  end
end
```
we could also do:
```elixir
  def authorize(%User{role: "admin"}, _), do: :ok

  def authorize(%User{account_id: account_id}, %CannedResponse{account_id: account_id}),
    do: :ok

  def authorize(_, _), do: {:error, :forbidden}
```
(which is basically just syntatic sugar for the case statement above)

a8t: thanks for this review, i appreciate it. don't worry about offending with nitpicks 🙂

PR #633, opened 2021-03-12
Add CRUD API boilerplate for canned response

jeepers3327:
### Description

Adds auto-generated files such as db migration, views, controller, context, and schema. Created unique index for name and account and implement unique_constraint in `canned_response` schema. 

Perform test including uniqueness of name  per account. I also added test factory for canned response for future usage.

### Issue

#623 

## Checklist

- [X] Ran `mix format`

### DIFF
--- a/lib/chat_api/canned_responses.ex
+++ b/lib/chat_api/canned_responses.ex
@@ -8,15 +8,7 @@ defmodule ChatApi.CannedResponses do
 
   alias ChatApi.CannedResponses.CannedResponse
 
-  @doc """
-  Returns the list of canned_responses.
 
-  ## Examples
-
-      iex> list_canned_responses()
-      [%CannedResponse{}, ...]
-
-  """
   def list_canned_responses do
     Repo.all(CannedResponse)
   end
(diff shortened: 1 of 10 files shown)


jeepers3327:
I'm not sure which PR would be accepted but I would appreciate a review @reichert621 so I could keep note of it in the future (I'm a junior dev).

lib/chat_api_web/controllers/canned_response_controller.ex
```diff
@@ -0,0 +1,45 @@
+defmodule ChatApiWeb.CannedResponseController do
+  use ChatApiWeb, :controller
+
+  alias ChatApi.CannedResponses
+  alias ChatApi.CannedResponses.CannedResponse
+
+  action_fallback ChatApiWeb.FallbackController
+
+  def index(conn, _params) do
```
reichert621, line 9: @jeepers3327 great work on all this! 

main feedback/thoughts:
- for all of these controller methods, make sure we use the `conn.assigns.current_user`'s information to filter the records we're querying
- for example, this `index` method as it is right now would fetch ALL records, rather than just the ones for the authenticated account
- also, for `create` and `update` methods, we need to make sure that the `account_id` always matches the currently authenticated user's account -- we can do this by doing a `Map.merge(params, %{"account_id" => conn.assigns.user.account_id})`, or raising a 403/forbidden error if someone attempts to pass in a different `account_id`

and if you don't mind, it would be great to add typespecs to these methods as well!

(let me know if any of that doesn't make sense or isn't clear)

reichert621 (changes requested):
great work so far! left some feedback

lib/chat_api_web/controllers/canned_response_controller.ex
jeepers3327: Finish adding filtering for the canned  response controller. I would like to ask if there's a need for another test for the recent code change.

Would like to ask for review or feedback @reichert621 .

Thank you.

lib/chat_api/canned_responses.ex
```diff
@@ -0,0 +1,49 @@
...
+  @spec update_canned_response(CannedResponse.t(), map()) ::
+          {:ok, CannedResponse.t()} | {:error, Ecto.Changeset.t()}
+  def update_canned_response(%CannedResponse{} = canned_response, attrs) do
+    canned_response
+    |> CannedResponse.changeset(attrs)
+    |> Repo.update()
+  end
+
+  @spec delete_canned_response(CannedResponse.t()) ::
+          {:ok, Tag.t()} | {:error, Ecto.Changeset.t()}
```
reichert621, line 40:
```suggestion
          {:ok, CannedResponse.t()} | {:error, Ecto.Changeset.t()}
```

test/chat_api/canned_responses_test.exs
```diff
@@ -0,0 +1,89 @@
...
+    @invalid_attrs %{content: nil, name: nil}
+
+    setup do
+      account = insert(:account)
+      canned_response = insert(:canned_response, account: account)
+
+      {:ok, account: account, canned_response: canned_response}
+    end
+
+    test "list_canned_responses/0 returns all canned_responses", %{
```
reichert621, line 22:
```suggestion
    test "list_canned_responses/1 returns all canned_responses", %{
```

PR #634, opened 2021-03-13
Add conversations.last_activity_at 

fmterrorf:
### Description

Create a new field `conversations.last_activity` that gets populated when a message is created

### Issue
#451 

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api/messages.ex
+++ b/lib/chat_api/messages.ex
@@ -6,6 +6,7 @@ defmodule ChatApi.Messages do
   import Ecto.Query, warn: false
 
   alias ChatApi.Repo
+  alias ChatApi.Conversations
   alias ChatApi.Messages.{Message, MessageFile}
 
   require Logger
@@ -50,9 +51,15 @@ defmodule ChatApi.Messages do
 
   @spec create_message(map()) :: {:ok, Message.t()} | {:error, Ecto.Changeset.t()}
   def create_message(attrs \\ %{}) do
-    %Message{}
-    |> Message.changeset(attrs)
-    |> Repo.insert()
+    with {:ok, message} <-
+           %Message{}
+           |> Message.changeset(attrs)
+           |> Repo.insert() do
+      Conversations.mark_activity(message.conversation_id)
+      {:ok, message}
+    else
+      error -> error
+    end
   end
 
   @spec create_and_fetch!(map()) :: Message.t() | {:error, Ecto.Changeset.t()}
--- a/lib/chat_api/messages.ex
+++ b/lib/chat_api/messages.ex
@@ -131,7 +131,8 @@ defmodule ChatApi.Messages do
   end
 
   defp after_message_created({:ok, message} = params) do
-    Workers.MessageCreatedActions.new(%{"id" => message.id})
+    %{"id" => message.id}
+    |> Workers.MessageCreatedActions.new()
     |> Oban.insert()
 
     params
--- a/lib/chat_api/messages.ex
+++ b/lib/chat_api/messages.ex
@@ -6,7 +6,7 @@ defmodule ChatApi.Messages do
   import Ecto.Query, warn: false
 
   alias ChatApi.Repo
-  alias ChatApi.Conversations
+  alias ChatApi.Workers
   alias ChatApi.Messages.{Message, MessageFile}
 
   require Logger
@@ -51,15 +51,10 @@ defmodule ChatApi.Messages do
 
   @spec create_message(map()) :: {:ok, Message.t()} | {:error, Ecto.Changeset.t()}
   def create_message(attrs \\ %{}) do
-    with {:ok, message} <-
-           %Message{}
-           |> Message.changeset(attrs)
-           |> Repo.insert() do
-      Conversations.mark_activity(message.conversation_id)
-      {:ok, message}
-    else
-      error -> error
-    end
+    %Message{}
+    |> Message.changeset(attrs)
+    |> Repo.insert()
+    |> after_message_created()
   end
 
   @spec create_and_fetch!(map()) :: Message.t() | {:error, Ecto.Changeset.t()}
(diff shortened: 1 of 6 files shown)


lib/chat_api/messages.ex
```diff
@@ -6,6 +6,7 @@ defmodule ChatApi.Messages do
   import Ecto.Query, warn: false
 
   alias ChatApi.Repo
+  alias ChatApi.Conversations
```
fmterrorf, line 9: I'm not too sure about aliasing `ChatApi.Conversations` here because that module uses `ChatApi.Messages` too. Feels like there's a circular dependency going on

priv/repo/migrations/20210312222205_add_conversations_last_activity_at.exs
```diff
@@ -0,0 +1,21 @@
...
+  def up do
+    alter table(:conversations) do
+      add(:last_activity_at, :utc_datetime)
+    end
+
+    execute("""
+    UPDATE conversations SET last_activity_at = subquery.recent_inserted_at
+    FROM (SELECT conversation_id, MAX(inserted_at) AS recent_inserted_at FROM messages GROUP BY conversation_id) AS subquery
+    WHERE subquery.conversation_id = conversations.id;
+    """)
```
fmterrorf, line 13: hopefully using raw sql is OK here :sweat_smile:

lib/chat_api/messages.ex
```diff
@@ -50,9 +51,15 @@ defmodule ChatApi.Messages do
...
   @spec create_message(map()) :: {:ok, Message.t()} | {:error, Ecto.Changeset.t()}
   def create_message(attrs \\ %{}) do
-    %Message{}
-    |> Message.changeset(attrs)
-    |> Repo.insert()
+    with {:ok, message} <-
+           %Message{}
+           |> Message.changeset(attrs)
+           |> Repo.insert() do
+      Conversations.mark_activity(message.conversation_id)
```
reichert621, line 58: hmmmm I feel a little FUD about including this as a side effect within the `create_message` method... I know it's a bit more repetitive but I slightly prefer invoking the `Conversations.mark_activity` method explicitly after wherever `create_message` is called 🤔 

I wonder if Phoenix has something similar to Rails' `after commit` hook (i.e. a hook that gets called after a new resource is successfully inserted into the db)... know anything like this?

reichert621 (changes requested):
thanks for tackling this! it should make paginating a lot easier :) 

just one concern about baking the `mark_activity` side effect into the `create_message` method... curious to hear your thoughts

lib/chat_api/messages.ex
fmterrorf: > I know it's a bit more repetitive but I slightly prefer invoking the Conversations.mark_activity method explicitly after wherever create_message is called 🤔

Yeah I had this feeling too. Like we're coupling Conversations and Messages, if I call `Conversations.mark_activity` here. I'm down to changing this so that we explicitly call mark_activity every time create_message is called.  

> I wonder if Phoenix has something similar to Rails' after commit hook (i.e. a hook that gets called after a new resource is successfully inserted into the db)... know anything like this?

From what I see from googling so far, doesn't seem like Ecto has that kind of callback. I'm looking at this reddit [thread](https://www.reddit.com/r/elixir/comments/5yo4ui/ecto_lack_of_callbacks_the_pain/)

fmterrorf: What do you think of having create_message publish an event somewhere. Then it'll get consumed by the app later, then we call `mark_activity`.

reichert621: I like that! right now we have a lot of repetition around callbacks when a new message is created, so setting up an event like that could be really useful for other things as well 👍 

mind explaining a bit how you're planning on implementing it, just so I have a better sense of what to expect?

fmterrorf: For sure. I'm thinking of using Oban. 
1. Create a worker called `ChatApi.Workers.NewMessageCreatedActions`
2. Every time a message gets called we enqueue a new job to that worker
3. The worker will be responsible for calling the callbacks that gets triggered when a new message is created. For example calling `Conversations.mark_activity`

reichert621: perfect! for now let's just start off with handling the `mark_activity` logic in that worker, we can handle refactoring some of the other logic later :+1:

fmterrorf: Sounds good :+1: I'll try to get these changes up tomorrow!

reichert621: great 😄

fmterrorf: pushed a new commit for this :pray:

lib/chat_api/messages.ex
```diff
@@ -127,4 +129,13 @@ defmodule ChatApi.Messages do
         dynamic
     end)
   end
+
+  defp after_message_created({:ok, message} = params) do
+    Workers.MessageCreatedActions.new(%{"id" => message.id})
```
reichert621, line 134: verrrry minor stylistic preference, but when i'm using pipes i prefer to start with the initial argument at the top like so:
```suggestion
    %{"id" => message.id}
    |> Workers.MessageCreatedActions.new()
```

priv/repo/migrations/20210312222205_add_conversations_last_activity_at.exs
reichert621: can't see why not 😛

reichert621 (approved):
awesome, thanks @fmterrorf! 🎉 

i'm gonna deploy to staging and play around with it a bit first before merging, but looks good :+1:

fmterrorf:
Once this gets merged, I have a branch that implements the paginate for this https://github.com/fmterrorf/papercups/compare/add_last_activity_to_conversations...fmterrorf:paginate_conversations

Issue #638, opened 2021-03-16
Improve security in conversation/message/customer controller show/update/delete methods


reichert621:
**Problem**
Right now any user with a valid auth token can retrieve any conversation/message/customer resource as long as they know the ID of the resource. We should return either a 404 (Not found) or 403 (Forbidden) error if someone tries to retrieve/update/delete a resource that does not have a matching `account_id`

**Solution**
In the `show`, `update`, and `delete` methods of the conversation/message/customer controllers, we can either:
- add a new method e.g. `Messages.get_message!(message_id, account_id)` method that filters by `account_id` in addition to the `message_id` and return a 404 error if not found
- or, using the result of `Messages.get_message!(message_id)`, check if the `message.account_id` matches the `conn.assigns.current_user.account_id`... if it doesn't, return a 403 (forbidden) error

**Describe alternatives you've considered**
I'm not sure which approach is better... on one hand, a 403 error is a bit more descriptive, but it makes it obvious when a user has "guessed" the ID of another account resource... so maybe a 404 is better? 🤷 

**Testing**
Please add/update tests to take these updates into account!

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

jeepers3327:
Can I work on this one?

a8t:
I would vote for 404. An unauthorized user doesn't need it to be a bit more descriptive, right?

I would also vote for the 2nd of your suggestions, doing the check in the controller. I think that it separates the concern of the controller (who knows about the interface and the current user etc) from the concern of the Messages context

@jeepers3327 you can look at my PR for that previous issue we both worked on for some work I already did on this front, including tests and a callback controller call. Or, since I'm halfway into this one, I can keep working on it

jeepers3327:
Right. Since you already started working on it @a8t , should be assigned to you. 
Thanks for informing me.

reichert621:
hey guys! @wmnnd is working on this one for now, but there are some other controllers that will need similar fixes (e.g. canned responses) -- i still need to create tickets for those as well but if you see any controllers with this same issue feel free to put up a PR @a8t @jeepers3327!

reichert621:
[...]

PR #639, opened 2021-03-16
Add authorization to ConversationController, CustomerController, MessageController

wmnnd:
Adds plug-based authorization for show/update/delete routes as suggested
in #638.
Also as suggested in #638, uses 404 status code when no authorized
resource was found.

### Issue
#638 

## Checklist

- [X] Ran `mix format`

### DIFF
--- a/lib/chat_api_web/controllers/customer_controller.ex
+++ b/lib/chat_api_web/controllers/customer_controller.ex
@@ -8,6 +8,25 @@ defmodule ChatApiWeb.CustomerController do
 
   action_fallback(ChatApiWeb.FallbackController)
 
+  plug :authorize when action in [:show, :update, :delete]
+
+  defp authorize(conn, _) do
+    id = conn.path_params["id"]
+
+    preloads =
+      (conn.path_params["expand"] || ["company", "tags"])
+      |> Enum.map(&String.to_existing_atom/1)
+      |> Enum.filter(&Customers.is_valid_association?/1)
+
+    with %{account_id: account_id} <- conn.assigns.current_user,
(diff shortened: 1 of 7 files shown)


wmnnd:
The scope of this PR is currently as described in #638, i.e. only implementing authorization for the `:show`/`:update`/`:delete` routes. Authorization should probably be added to all other routes of the affected controllers as well.

lib/chat_api_web/controllers/customer_controller.ex
```diff
@@ -8,6 +8,25 @@ defmodule ChatApiWeb.CustomerController do
 
   action_fallback(ChatApiWeb.FallbackController)
 
+  plug :authorize when action in [:show, :update, :delete]
+
+  defp authorize(conn, _) do
+    id = conn.path_params["id"]
+
+    preloads =
+      (conn.path_params["expand"] || ["company", "tags"])
```
reichert621, line 17: hmm i wonder if we can remove the "company"/"tags" defaults... probably safer to keep them in for now 👍 (just commenting so i don't forget)

reichert621 (approved):
thanks @wmnnd! cleans things up nicely 🎉 🎉 🎉

lib/chat_api_web/controllers/customer_controller.ex
wmnnd: It’s from the old `get_customer!/1` and removing this default seems to break three or four tests.

reichert621: makes sense!

PR #645, opened 2021-03-16
Add test watcher package

a8t:
### Description

Add a test watcher mix package: https://github.com/lpil/mix-test.watch

## Checklist

- [x] Everything passes when running `mix test` 😏
- [x] Ran `mix format`


reichert621:
nice, seems useful!

PR #648, opened 2021-03-17
Paginate conversations

fmterrorf:
### Description
- Created Conversations.list_conversations_by_account_paginated
- Installed this lib https://github.com/duffelhq/paginator to help us with pagination

### Issue
#451

Also this PR is a continuation of this https://github.com/papercups-io/papercups/pull/634

## Checklist

- [x] Ran `mix format`


test/chat_api/conversations_test.exs
```diff
@@ -709,6 +709,51 @@ defmodule ChatApi.ConversationsTest do
...
+      %{metadata: metadata1, entries: entries1} =
+        Conversations.list_conversations_by_account_paginated(account.id, filter)
+
+      assert Enum.map(entries1, & &1.id) == Enum.map(third_batch, & &1.id)
+
+      %{metadata: metadata2, entries: entries2} =
+        Conversations.list_conversations_by_account_paginated(
+          account.id,
+          Map.merge(filter, %{
+            "after" => metadata1.after
```
fmterrorf, line 737: `after` here is an encoded representation of the cursor pagination. The frontend needs to get a hold of this value, so that it can pass it on the request and get the next set of result

a8t (commented):
Hey, I learned a good bit from another one of your PRs! I have two questions:

- why did you choose to do pagination in the context rather than in the controllers?
- would it be useful to add a controller test?

Thanks again!

lib/chat_api/conversations.ex
```diff
@@ -23,6 +23,19 @@ defmodule ChatApi.Conversations do
     |> Repo.all()
   end
 
+  @spec list_conversations_by_account_paginated(binary(), map()) :: Paginator.Page.t()
+  def list_conversations_by_account_paginated(account_id, filters \\ %{}) do
+    Conversation
+    |> where(account_id: ^account_id)
+    |> where(^filter_where(filters))
+    |> where([c], is_nil(c.archived_at))
+    |> order_by(desc: :last_activity_at, desc: :id)
```
a8t, line 32: I'm curious, why is the sort different than both the v1 above for the v2 below? I assume this is what prevents you from duplicating the code. Is this a sorting specific thing?

fmterrorf:
Hi @a8t 
>  * why did you choose to do pagination in the context rather than in the controllers?

The requirement for this issue is to only create the context layer https://github.com/papercups-io/papercups/issues/451#issue-758058239

>   * would it be useful to add a controller test?

Yeah i think we could add controller tests once we start hooking it up to a controller

lib/chat_api/conversations.ex
fmterrorf: Yeah this PR is a continuation of this other PR https://github.com/papercups-io/papercups/pull/634
We decided to create a `last_activity_at` field in the `conversations` table so that we don't have a lateral join to the `message` field to sort which conversation has the most recent message, which the method below uses

a8t:
Makes sense, thanks again :)

lib/chat_api/conversations.ex
reichert621, line 32: just curious, what is the `desc: :id` for? is it just the secondary ordering? if that's the case can we use `inserted_at` instead?

lib/chat_api/conversations.ex
```diff
@@ -23,6 +23,19 @@ defmodule ChatApi.Conversations do
...
+    |> preload([:customer, messages: [:attachments, :customer, user: :profile]])
+    |> Repo.paginate_with_cursor(
+      filter_pagination([cursor_fields: [last_activity_at: :desc, id: :desc]], filters)
```
reichert621, line 35: instead of lumping the pagination options in with `filters`, can we put it in a 3rd argument for now? then we can probably just remove the `filter_pagination` method right?

test/chat_api/conversations_test.exs
```diff
@@ -709,6 +709,51 @@ defmodule ChatApi.ConversationsTest do
...
   end
 
+  describe "list_conversations_by_account_paginated/1" do
+    test "sorts the conversations by most recent activity" do
+      account = insert(:account)
+      filter = %{"limit" => 5}
+      base_last_activity_at = DateTime.from_naive!(~N[2020-12-01 00:00:00], "Etc/UTC")
+
+      [third_batch, second_batch, first_batch] =
+        15..1
```
reichert621, line 719: maybe i'm reading this incorrectly but just curious why not `1..15` and then use `first_batch, second_batch, third_batch` above?

reichert621 (changes requested):
just a few small comments! looks good though, much better now that we have the `last_activity_at` field :)

lib/chat_api/conversations.ex
fmterrorf: yeah I decided to sort using the `id` field so that we'll have a more deterministic results. In the example in this [page](https://use-the-index-luke.com/sql/partial-results/fetch-next-page), the used the `sale_id`, and the paginator doc also uses it for their [example](https://github.com/duffelhq/paginator#usage)

The paginator also mentioned
> If you want to reap all the benefits of this method it is better that you create indexes on the columns you are using as cursor fields.


The ID field also already indexed, since it's a primary key. I'd still need to add an index for `last_activity_at`

test/chat_api/conversations_test.exs
fmterrorf: I used 15...1 because I want the results to have high to low `last_activity_value`. That way the first page will be the ones with the most recent activity, then the second page will have results with 2nd most recent activity

I guess I can change the name from `third_batch, second_batch, first_batch` to `first_batch, second_batch, third_batch`

lib/chat_api/conversations.ex
reichert621: got it, makes sense!

test/chat_api/conversations_test.exs
reichert621: ah i see -- yeah maybe reverse the naming? that would be a little less confusing (otherwise it just feels like everything is flipped)

lib/chat_api/conversations.ex
fmterrorf: Sounds good :+1: I moved it to the third arg

fmterrorf:
Add index for `last_actiivity_at`

reichert621 (approved):
nice!

Issue #643, opened 2021-03-16
Improve security in `TagController` show/update/delete methods


reichert621:
**Problem**
Right now any user with a valid auth token can retrieve any `tag` resource as long as they know the ID of the resource. We should return a 404 (Not found) error if someone tries to retrieve/update/delete a resource that does not have a matching `account_id`

**Solution**
In the `show`, `update`, and `delete` methods of the tag controller, let's add an `authorize` plug (middleware) to handle this.

Check out https://github.com/papercups-io/papercups/pull/639 for an example

**Describe alternatives you've considered**
We could also just include the `account_id` in the query when we fetch the resource 🤷 

**Testing**
Please add/update tests to take these updates into account!

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

a8t:
I can do this one too, assuming my other PR is approved.

Issue #641, opened 2021-03-16
Improve security in `CompanyController` show/update/delete methods


reichert621:
**Problem**
Right now any user with a valid auth token can retrieve any `company` resource as long as they know the ID of the resource. We should return a 404 (Not found) error if someone tries to retrieve/update/delete a resource that does not have a matching `account_id`

**Solution**
In the `show`, `update`, and `delete` methods of the company controller, let's add an `authorize` plug (middleware) to handle this.

Check out https://github.com/papercups-io/papercups/pull/639 for an example

**Describe alternatives you've considered**
We could also just include the `account_id` in the query when we fetch the resource 🤷 

**Testing**
Please add/update tests to take these updates into account!

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

ghsyeung:
Hi, I really like this project :+1: and looking forward to using it at our company.
I have some questions after looking at #639.

First of all I don't have experience in Elixir + Phoenix so please point out if my understanding is incorrect.

My understanding is that `plug` acts like middleware that's called prior to the controller action and the `conn` object should be returned or `halt` should be called (e.g. if the current user doesn't have sufficient privileges to call the controller action.)

In #639, I believe the logic works as follow.

1. a private method `:authorize` is called when the action matches against list
    -  if user is not authorized
       1. a response (404 - Bad Request) is sent back to the user (through the `FallbackController`)
       2. call `halt` to terminate the request

I have some suggestions that may improve security as well as maintainability.

By calling `:authorize` only when an action matches a list, it means actions are **default to open** (i.e. they don't require any authorization). So if someone accidentally forgets to add a protected action to the list, the protected resource would be accessible by anyone.

The alternative approach would be **default to close**. This means the `:authorize` plug is always called. If the action isn't specified, it would fail the request. For example,

```elixir
# not sure if this is the correct syntax
plug Auth.authorize, 
    %{ 
       show: open_to_all, 
       update: can_access_company, 
       delete: can_access_company,       
     }

# this can be defined in the Auth module
def :open_to_all(conn, _) do
    # do nothing, just return `conn`
    conn
end

defp :can_access_company(conn, _) do
[...]

PR #649, opened 2021-03-17
Add authorize plug to tag and company controllers

a8t:
### Description

Adds a plug for authorizing some Tag and Company endpoints by default.

Also has tests for `:show`, `:update`, and `:delete` cases.

Initially I just did the `Tag` controller, but it was straight up a copy+paste job to cover the `Company` controller.

## Issues

https://github.com/papercups-io/papercups/issues/641
https://github.com/papercups-io/papercups/issues/643

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api_web/controllers/company_controller.ex
+++ b/lib/chat_api_web/controllers/company_controller.ex
@@ -6,7 +6,7 @@ defmodule ChatApiWeb.CompanyController do
 
   action_fallback ChatApiWeb.FallbackController
 
-  plug :authorize when action not in [:index, :create]
+  plug :authorize when action in [:show, :update, :delete]
 
   defp authorize(conn, _) do
     id = conn.path_params["id"]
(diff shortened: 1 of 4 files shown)


a8t:
@reichert621

lib/chat_api_web/controllers/company_controller.ex
```diff
@@ -6,6 +6,19 @@ defmodule ChatApiWeb.CompanyController do
 
   action_fallback ChatApiWeb.FallbackController
 
+  plug :authorize when action not in [:index, :create]
```
reichert621, line 9: let's still with `in` vs `not in` for now, i think it makes it a bit more obvious
```suggestion
  plug :authorize when action in [:show, :update, :delete]
```

lib/chat_api_web/controllers/tag_controller.ex
```diff
@@ -6,6 +6,19 @@ defmodule ChatApiWeb.TagController do
 
   action_fallback ChatApiWeb.FallbackController
 
+  plug :authorize when action not in [:index, :create]
```
reichert621, line 9: same here
```suggestion
  plug :authorize when action in [:show, :update, :delete]
```

reichert621 (commented):
thanks for this :) just one small nitpick

PR #657, opened 2021-03-18
Add Bodyguard library to canned responses

a8t:
### Description

Switches hand-rolled auth in controller to [Bodyguard](https://hexdocs.pm/bodyguard/readme.html) auth, with auth policy defined in context. 

### Issue

As discussed in https://github.com/papercups-io/papercups/issues/641

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api/canned_responses.ex
+++ b/lib/chat_api/canned_responses.ex
@@ -1,4 +1,6 @@
 defmodule ChatApi.CannedResponses do
+  defdelegate authorize(action, user, params), to: ChatApi.CannedResponses.Policy
+
   @moduledoc """
   The CannedResponses context.
   """
@@ -46,4 +48,25 @@ defmodule ChatApi.CannedResponses do
   def change_canned_response(%CannedResponse{} = canned_response, attrs \\ %{}) do
     CannedResponse.changeset(canned_response, attrs)
   end
+end
+
+defmodule ChatApi.CannedResponses.Policy do
+  @behaviour Bodyguard.Policy
+
+  # Require account_id check for getting, updating, deleting
+  def authorize(
+        action,
+        %{account_id: account_id} = _user,
+        %{account_id: account_id} = _canned_response
+      )
+      when action in [:get_canned_response!, :update_canned_response, :delete_canned_response],
+      do: :ok
+
+  # Don't require authorization check when listing or creating
+  def authorize(action, _, _)
+      when action in [:list_canned_responses, :create_canned_response],
+      do: :ok
+
+  # Catch-all: deny everything else
+  def authorize(_, _, _), do: {:error, :not_found}
 end
--- a/lib/chat_api_web/controllers/canned_response_controller.ex
+++ b/lib/chat_api_web/controllers/canned_response_controller.ex
@@ -6,9 +6,10 @@ defmodule ChatApiWeb.CannedResponseController do
 
   action_fallback ChatApiWeb.FallbackController
 
-  plug :get_canned_response when action in [:show, :update, :delete]
+  @unauthorized_actions [:index, :create]
 
-  # Do the check
+  plug :get_canned_response when action not in @unauthorized_actions
+
   plug Bodyguard.Plug.Authorize,
        [
          policy: CannedResponses.Policy,
@@ -17,7 +18,7 @@ defmodule ChatApiWeb.CannedResponseController do
          params: {__MODULE__, :extract_canned_response},
          fallback: ChatApiWeb.FallbackController
        ]
-       when action in [:show, :update, :delete]
+       when action not in @unauthorized_actions
 
   # Helper for the Authorize plug
   def current_user(conn), do: conn.assigns.current_user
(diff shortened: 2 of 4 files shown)


lib/chat_api/canned_responses.ex
```diff
@@ -1,4 +1,6 @@
 defmodule ChatApi.CannedResponses do
+  defdelegate authorize(action, user, params), to: ChatApi.CannedResponses.Policy
```
a8t, line 2: You can put the auth stuff directly in the context module, but I didn't like it.

lib/chat_api_web/controllers/canned_response_controller.ex
```diff
@@ -42,25 +29,34 @@ defmodule ChatApiWeb.CannedResponseController do
   end
 
   @spec show(Plug.Conn.t(), map()) :: Plug.Conn.t()
-  def show(conn, _params) do
-    render(conn, "show.json", canned_response: conn.assigns.current_canned_response)
+  def show(conn, %{"id" => id}) do
```
a8t, line 32: I do like that this makes us be more explicit about needing `id` as a param.

mix.exs
```diff
@@ -75,7 +75,8 @@ defmodule ChatApi.MixProject do
       {:tzdata, "~> 1.0.5"},
       {:scrivener_ecto, "~> 2.0"},
       {:floki, "~> 0.30.0"},
-      {:mix_test_watch, "~> 1.0", only: :dev, runtime: false}
+      {:mix_test_watch, "~> 1.0", only: :dev, runtime: false},
+      {:bodyguard, "~> 2.4"}
```
a8t, line 79: I'm thinking maybe these should be alphabetical. Sorry for bikeshedding 😛

a8t:
@reichert621

lib/chat_api_web/controllers/canned_response_controller.ex
```diff
@@ -6,17 +6,27 @@ defmodule ChatApiWeb.CannedResponseController do
...
-    with %{account_id: account_id} <- conn.assigns.current_user,
-         canned_response = %{account_id: ^account_id} <- CannedResponses.get_canned_response!(id) do
-      assign(conn, :current_canned_response, canned_response)
-    else
-      _ -> ChatApiWeb.FallbackController.call(conn, {:error, :not_found}) |> halt()
-    end
+  plug Bodyguard.Plug.Authorize,
+       [
+         policy: CannedResponses.Policy,
+         action: :get_canned_response!,
```
ghsyeung, line 16: I don't think that's the correct way to set the `action` provider.

https://hexdocs.pm/bodyguard/Bodyguard.Plug.Authorize.html#module-option-getters

https://github.com/schrockwell/bodyguard/blob/develop/lib/bodyguard/plug/authorize.ex#L118

It's basically call `action` on `conn`, so if you use `{Phoenix.Controller, :action_name}`, or `&action_name/1`, then you shouldn't need to specify the guard on line 21. And you should be able to use action names like `:index`, `:show`, etc inside the policy.

a8t: Yup, I know. That's why I don't like it. I don't think the policy should know or care about controller actions

ghsyeung: Ah! I see what you mean. I guess the decision is whether you authorize in controller (outside context) or within context. 

From my prior understanding of the other PR, authorization was mainly considered at the controller level. And I see how Bodyguard's examples ties `Policy` modules to the context.

If Policy ties to the context (i.e. specific rules by context operations), then using `permit` would be more natural (unless you create some time of lookup of `action -> [context operations]`). 
- Pros: You can control authorization at a fine-grained level.
- Cons: This can be error prone because you can call `permit` with incorrect arguments or not call it at all.

On the other hand, if the Policy ties to the controller (i.e. rules are defined by controller actions).
- Pros: `Plug` approach would be more natural
- Cons: authorization is defined by controller actions

Fundamentally, from what I read, I don't think Bodyguard's logic cares whether the `Policy` is defined in the context or controller. Would it be too far off the convention (Phoenix) to define the `Policy` module in the controller?

Ultimately, there's a design decision to be made on whether authorization should be done at the context level or the controller level. (https://hexdocs.pm/bodyguard/readme.html#where-should-i-perform-checks)

reichert621: honestly to me it seems to make more sense for the authorization to be done at the controller level. it seems like just because the bodyguard logic _can_ be set at the context level doesn't mean it _has_ to be (as noted in the link you shared above)

anyway, as interesting as this is, it feels like we're starting to yak shave a bit 😛 i'd love to see how this looks like the authorization handled at the controller level with plugs as a proof of concept, but in the meantime i think we're just going to stick with the existing solution we have (just a simple `authorize` plug explicitly set on specified controller actions)

thanks for indulging me with this discussion so far @a8t @ghsyeung! (tl;dr: i'm leaning toward eventually using bodyguard at the controller level with plugs, but for this week it's not a priority)

a8t: Sounds good, I think we are in agreement about the design and it would be a good future revision after sleeping on it for a few cycles.

(Regarding authorize plug, I also had a PR for that https://github.com/papercups-io/papercups/pull/649)

PR #658, opened 2021-03-18
Set frontend env variables from server

reichert621:
### Description

Sets `window.__ENV__` on server so we don't require a complete redeploy whenever we need to update these env variables on the frontend.

### Issue

https://create-react-app.dev/docs/title-and-meta-tags/#injecting-data-from-the-server-into-the-page

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api_web/controllers/page_controller.ex
+++ b/lib/chat_api_web/controllers/page_controller.ex
@@ -3,6 +3,35 @@ defmodule ChatApiWeb.PageController do
 
   @spec index(Plug.Conn.t(), map()) :: Plug.Conn.t()
   def index(conn, _params) do
-    html(conn, File.read!("./priv/static/index.html"))
+    file =
+      "./priv/static/index.html"
+      |> File.read!()
+      |> String.replace(
+        "__SERVER_ENV_DATA__",
+        Jason.encode!(server_env_data(), escape: :html_safe)
+      )
+
+    html(conn, file)
+  end
+
+  defp server_env_data() do
+    %{
+      REACT_APP_SENTRY_DSN: System.get_env("REACT_APP_SENTRY_DSN"),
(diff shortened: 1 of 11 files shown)


lib/chat_api_web/controllers/page_controller.ex
```diff
@@ -3,6 +3,35 @@ defmodule ChatApiWeb.PageController do
...
+        "__SERVER_ENV_DATA__",
+        Jason.encode!(server_env_data(), escape: :html_safe)
+      )
+
+    html(conn, file)
+  end
+
+  defp server_env_data() do
+    %{
+      REACT_APP_SENTRY_DSN: System.get_env("REACT_APP_SENTRY_DSN"),
```
fmterrorf, line 19: Can this configs live in `config/config.exs`? file. https://hexdocs.pm/elixir/master/Config.html. Then we can use Application.fetch_env! to fetch the values for it

reichert621: good idea :+1:

reichert621: curious though, what's the benefit of that over this approach ?

fmterrorf: No tangible benefits I think 😅  other than having all the configs in one place

fmterrorf: I guess this also makes it so that you can have different config values per env. Like you can override configs for `config/dev.exs`, `config/test.exs` etc.

reichert621: makes sense! I'm going to merge this as is for now so we get the fixes we need in place, but will revisit cleaning this up soon :)

PR #664, opened 2021-03-19
Set up boilerplate for Twilio integration

reichert621:
### Description

This PR will set up the boilerplate logic for a basic Twilio integration:
- [x] Add `ex_twilio` dependency
- [x] Add `ex_twilio` config
- [ ] Set up the webhook URL to receive incoming SMS messages

_**In separate PR:**_
- When agent replies from dashboard/Slack/etc AND conversation `source: "sms"` AND customer has valid phone number, send SMS message to customer via Twilio API using account's `twilio_authorization` info

### Issue

TODO: write up ticket(s)

## Checklist

- [ ] Ran `mix format`


reichert621:
closing since we don't actually want to configure twilio at the application level... we'll want users to provide their own API information if they want to use it

moving to https://github.com/papercups-io/papercups/pull/671

PR #666, opened 2021-03-19
Alphabetize Mix deps

a8t:
### Description

Here's a Hacktoberfest-tier PR...

Alphabetize Mix dependencies. It's not that important, but why not? ¯\\\_(ツ)_/¯

- [x] Ran `mix format`

(p.s. shout-out to whatever VS Code line sorter extension I have installed)


reichert621 (commented):
seems reasonable! i wonder if there's a way to configure `mix format` (or some other pre-commit hook) to do this automatically 🤔

a8t:
Short answer is "no" re: `mix format`. Slightly longer answer is that we can probably write a script that uses the [Code](https://hexdocs.pm/elixir/Code.html) module to parse the AST and maybe sort the Mix file, but that seems... kinda bad, kinda cool 🤔

Divspy22d:
Good and awesome

Divspy22d:
Screenshot_20210303-220726

Divspy22d:
Nice update

PR #671, opened 2021-03-22
Set up Twilio authorization flow

reichert621:
### Description

This PR will set up the boilerplate logic for storing Twilio authorizations:
- [x] Add the `twilio_authorizations` table with `auth_token`, `account_sid`, `from_phone_number`, `account_id`, etc
- [x] Add CRUD methods for managing these authorizations
- [x] Set up the placeholder webhook URL to receive incoming SMS messages

_**In separate PR:**_
- When agent replies from dashboard/Slack/etc AND conversation `source: "sms"` AND customer has valid phone number, send SMS message to customer via Twilio API using account's `twilio_authorization` info

### Issue

Handles https://github.com/papercups-io/papercups/issues/674

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/lib/chat_api/messages/notification.ex
+++ b/lib/chat_api/messages/notification.ex
@@ -54,6 +54,22 @@ defmodule ChatApi.Messages.Notification do
     message
   end
 
+  def notify(%Message{private: false} = message, :sms, _opts) do
+    Logger.info("Sending message notification: :sms")
+
+    Task.start(fn ->
+      # TODO: implement me!
+      #
+      # First, check if message parent conversation has `source: "sms"`
+      # Also check that the conversation customer has a valid phone number
+      # If true, use the message account ID to check for a twilio_authorization record
+      # If one exists, use it to send a message to the customer's phone number
+      nil
+    end)
+
+    message
+  end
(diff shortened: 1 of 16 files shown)


lib/chat_api/messages/notification.ex
```diff
@@ -54,6 +54,22 @@ defmodule ChatApi.Messages.Notification do
     message
   end
 
+  def notify(%Message{private: false} = message, :sms, _opts) do
+    Logger.info("Sending message notification: :sms")
+
+    Task.start(fn ->
```
fmterrorf, line 60: Just wondering if if we should make this task [supervised](https://hexdocs.pm/elixir/master/Task.html#module-dynamically-supervised-tasks). So that we can make the node wait for pending tasks before it shuts down. [Task.start](https://hexdocs.pm/elixir/master/Task.html#start/1) wont wait

reichert621: yeah that could be a good idea! i've mostly just used `Task.start` in the meantime for simplicity, so if we can make this more robust with supervision that would be great :+1:

Issue #332, opened 2020-10-19
Paginate conversations queries in each Inbox tab


cheeseblubber:
**Is your feature request related to a problem? Please describe.**
Currently we fetch the full list of conversations in each Inbox tab. For users with a lot of conversations it can be slow to load.

**Describe the solution you'd like**
* We should be paginating the conversation and fetch the first 50 conversations. 
* Should be ordered by most recently sent message
* When the user scrolls to the bottom it should fetch the next 50 conversations, or display a button to load more.

**Additional context**
Right now we are listening to all conversations that are being loaded we should make sure that we're listening not just the conversations that are loaded with the pagination change.

Related to https://github.com/papercups-io/papercups/issues/132

PR #679, opened 2021-03-23
Set up pagination for conversations in dashboard

reichert621:
### Description

Set up pagination for conversations in dashboard

### Issue

https://github.com/papercups-io/papercups/issues/332

### Screenshots

load-more-conversations

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/assets/src/components/conversations/ConversationsDashboard.tsx
+++ b/assets/src/components/conversations/ConversationsDashboard.tsx
@@ -3,7 +3,8 @@ import {Box} from 'theme-ui';
 import qs from 'query-string';
 import {colors, Layout, notification, Sider, Title} from '../common';
 import {sleep} from '../../utils';
-import {Account, Message} from '../../types';
+import {ConversationsListResponse, PaginationOptions} from '../../api';
+import {Account, Conversation, Message} from '../../types';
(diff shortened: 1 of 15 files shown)


assets/src/components/conversations/ConversationsDashboard.tsx
```diff
@@ -285,9 +314,13 @@ class ConversationsDashboard extends React.Component<Props, State> {
   };
 
   render() {
-    const {selectedConversationId, closing = []} = this.state;
+    const {selectedConversationId, pagination = {}, closing = []} = this.state;
     const {title, conversationIds = []} = this.props;
     const loading = this.props.loading || this.state.loading;
+    const hasMoreConversations =
```
fmterrorf, line 320: Can check if `pagination.after` is not null, and use that to indicate if there's more conversations?

reichert621: i renamed `after`/`before` to `next`/`previous`, so we do the check for `pagination.next` below :+1:

(there's an edge case where we've already loaded all the conversations into the cache, which is why we also check that `conversationIds.length < pagination.total`)

Issue #678, opened 2021-03-23
Twilio integration v1


reichert621:
**Problem**
Users want to be able to receive messages from their customers via SMS (and eventually WhatsApp).

The easiest way to get an MVP up and running is to use Twilio.

**Solution**
Papercups users should be able to:
- Set up a Twilio authorization with their own API keys and phone number
- Point their Twilio account at a Papercups webhook URL to handle incoming SMS messages
- New messages will come in as new conversations, and followup messages will be handled in the same conversation

Tasks:
- Set up the Twilio authorization flow (https://github.com/papercups-io/papercups/pull/671)
- Set up the Twilio webhook handler (https://github.com/papercups-io/papercups/issues/675)
- Handle replies from Papercups being sent back via SMS (https://github.com/papercups-io/papercups/issues/676)
- Add documentation (https://github.com/papercups-io/papercups/issues/677)

**Additional context**
We'll use https://github.com/papercups-io/papercups/tree/twilio-integration-v1 as a feature branch for this

**Testing**
This will require a fair bit of manual testing, plus automated tests for the webhook handler in particular.

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

a8t:
I'm in! @fmterrorf how did you want to split it up? Maybe I can do #676

fmterrorf:
@a8t i'll have a look into details in a bit. Thinking of picking up #675

reichert621:
@a8t @fmterrorf let me know if you need any help getting set up with Twilio!

PR #680, opened 2021-03-24
Twilio integration v1 feature

reichert621:
This is the feature branch for the Twilio integration v1: https://github.com/papercups-io/papercups/issues/678

| Phone  | Dashboard | Slack |
| ------------- | ------------- | ------------- |
| Screen Shot 2021-04-02 at 10 19 25 AM | Screen Shot 2021-04-02 at 10 19 24 AM | Screen Shot 2021-04-02 at 10 19 33 AM |

### DIFF
--- a/lib/chat_api/twilio/notification.ex
+++ b/lib/chat_api/twilio/notification.ex
@@ -10,6 +10,7 @@ defmodule ChatApi.Twilio.Notification do
     Conversations.Conversation,
     Customers.Customer,
     Twilio,
+    Twilio.TwilioAuthorization,
     Messages.Message
   }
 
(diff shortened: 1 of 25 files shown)


lib/chat_api/twilio/notification.ex
```diff
@@ -10,23 +10,26 @@ defmodule ChatApi.Twilio.Notification do
...
+    Twilio.TwilioAuthorization,
     Messages.Message
   }
 
   @spec notify_sms(ChatApi.Messages.Message.t()) :: {:error, any} | {:ok, Tesla.Env.t()}
   def notify_sms(%Message{
         conversation_id: conversation_id,
         body: body,
-        account_id: account_id,
-        customer: %Customer{phone: customer_phone}
```
a8t: just curious - looks like most of the refactor here was in order to move this into the `with`. Is that becaue we can't be sure that the `customer` is preloaded? @reichert621

reichert621: in this case, the message is coming from the agent/user, not the customer -- so `customer` here will be `nil`... we need to get the `customer` from the message's `conversation` instead :)

reichert621:
thanks for the help on this @a8t @fmterrorf! 🎉 🎉 🎉

PR #681, opened 2021-03-24
Change Datetime calls to specify milliseconds

a8t:
### Description

Up to you if you think this is important, but I was bothered while writing some other tests based on the ones here. 

`:timer.hours/1` uses ms, but `DateTime.add/3` uses seconds by default unless you specify. The tests that I borrowed from didn't specify, so it led me to write incorrect tests. This should prevent that slight confusion in future.

## Checklist

- [x] Ran `mix format`


reichert621 (approved):
thanks @a8t!

PR #409, opened 2020-11-22
Upgrade to Elixir v1.11

reichert621:
### Description

https://elixir-lang.org/blog/2020/10/06/elixir-v1-11-0-released/

### Issue

Handles #278

## Checklist

- [x] Everything passes when running `mix test`
- [x] Ran `mix format`
- [x] No frontend compilation warnings

Issue #278, opened 2020-10-06
Upgrade to Elixir v1.11


reichert621:
This last version offers some nice updates: https://elixir-lang.org/blog/2020/10/06/elixir-v1-11-0-released/

Let's upgrade if it doesn't break anything :)

PR #687, opened 2021-03-25
Elixir upgrade

cheeseblubber:
### Description

Upgrades to using latest elixir version. Migrated to use runtime.exs instead of release.exs

### Issue

https://github.com/papercups-io/papercups/pull/409

https://github.com/papercups-io/papercups/issues/278

### Screenshots

## Checklist

- [x] Ran `mix format`


config/runtime.exs
```diff
@@ -0,0 +1,137 @@
...
+    secret_key_base: secret_key_base,
+    server: true,
+    check_origin: false
+
+  if https do
+    config :chat_api, ChatApiWeb.Endpoint,
+      https: [
+        port: 443,
+        cipher_suite: :strong,
+        otp_app: :hello,
```
reichert621, line 71: should this be `:chat_api`?

cheeseblubber: Good catch fixed it

Issue #676, opened 2021-03-23
[Twilio Integration v1] Send replies via SMS


reichert621:
**Problem**
When a user/agent replies to a conversation that was started over SMS, we want to make sure the message gets sent back via SMS (with Twilio)

**Solution**
When a user/agent replies from dashboard/Slack/etc:
- AND conversation `source: "sms"` 
- AND customer has valid phone number

...send SMS message to customer via Twilio API using account's `twilio_authorization` info.

**Additional context**
https://github.com/papercups-io/papercups/issues/678

The logic should live around here: https://github.com/papercups-io/papercups/pull/671/files#diff-37c06520fa000216238485bb22639f56ecf7b5be33477500ac4b798e0a561e5bR61

**Testing**
This may require some manual testing, but please add some automated tests as well!

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

a8t:
I can look into this

a8t:
Can you define 'valid phone number'? is that something Twilio defines?

a8t:
nts - https://support.twilio.com/hc/en-us/articles/223183668-Can-I-Test-Twilio-s-REST-API-Within-my-Application-Without-Actually-Making-the-Request-

PR #689, opened 2021-03-26
Send SMS replies via Twilio 

a8t:
### Description

When a sending a message, include sending an SMS if `source: sms`, valid phone #, and twilio account exists

### Issue

#676 

## Checklist

- [x] Ran `mix format`


a8t:
@reichert621 - What do you think is the best way to validate phone #s? I can look for regexes tomorrow... but that seems probably wrong.

What do you think of the big fat `with` statement?

PR #693, opened 2021-03-27
Add Twilio webhook handler

### LINKED ISSUE #675: [Twilio Integration v1] Set up webhook handler
**Problem**
When a user with a Twilio authorization (from https://github.com/papercups-io/papercups/issues/674) receives a new message, Twilio sends the data to a webhook URL (that is specified within the Twilio account).

When a new message comes in, we have to handle it appropriately (i.e. create a new conversation in Papercups, or add the message to an existing conversation, etc)

**Solution**

**Additional context**
https://github.com/papercups-io/papercups/issues/678

You will need to set up a (free) Twilio account and set up the webhook URL in https://www.twilio.com/console/sms/services (create messaging service, and then find "Integration" in side menu)

The logic for this will live here: https://github.com/papercups-io/papercups/pull/671/files#diff-08ff735cf4232fafb4171ebd89cfe3bf9af24e447356da9aaefc5d4f4663229aR65

**Testing**
Please write automated tests sending sample webhook events to the `/api/twilio/webhook` endpoint.

fmterrorf:
### Description
-  Added Twilio webhook handler

### Issue
Fixes #675 

### Screenshot
Here's a screenshot of a new conversation
twilio-test
I used ngrok to test this locally

## Checklist

- [x] Ran `mix format`


test/chat_api_web/controllers/twilio_controller_test.exs
```diff
@@ -0,0 +1,97 @@
...
+    @request_body %{
+      "AccountSid" => "1234",
+      "To" => "1234",
+      "From" => "1234",
+      "Body" => "body"
+    }
+
+    test "returns 200 when message is successfuly created",
+         %{authed_conn: authed_conn} do
+      with_mocks([
```
fmterrorf, line 33: I've decided to use mocks when testing in controller so that I don't have to insert actual data to the db. It seems like this makes sense because it separates the concerns between the controller vs context. The context should do all the business logic and the controller is responsible the web layer.

lib/chat_api_web/controllers/twilio_controller.ex
```diff
@@ -60,20 +61,35 @@ defmodule ChatApiWeb.TwilioController do
...
+           Messages.create_message(%{
+             body: body,
+             account_id: account_id,
+             customer_id: customer.id,
+             conversation_id: conversation.id
+           }) do
+      send_resp(conn, 200, "")
+    else
+      nil ->
+        send_resp(conn, 404, "Twilio account not found")
```
fmterrorf, line 87: Not sure if twilio should get a 404 response if the receiving number does not exist :thinking:

fmterrorf:
cc @reichert621

lib/chat_api/twilio.ex
```diff
@@ -82,6 +84,57 @@ defmodule ChatApi.Twilio do
     |> Repo.one()
   end
 
+  @spec find_or_create_customer_and_conversation(String.t(), String.t()) ::
+          {:ok, Customers.Customer.t(), Conversations.Conversation.t()}
+          | {:error, Ecto.Changeset.t()}
+  def find_or_create_customer_and_conversation(account_id, phone) do
```
fmterrorf, line 90: Im not sure if this should live here. I can see this being used for other integration from different sources.

a8t: I agree, I think it should live in conversations (or customers) but probably not here

a8t:
I noticed we changed some of the same files. I'm fine to handle conflicts since my PR is newer!

lib/chat_api_web/controllers/twilio_controller.ex
a8t: Is there some way to throw away requests to this that don't come from Twilio?

fmterrorf: I think we can just send 200 OK response, and have a warning log that includes a message that a twilio account is not found

a8t: Hmmm true, esp cuz I don't think twilio cares at all what we do with their webhooks 🤔

a8t:
Rebase troubles? 😛

fmterrorf:
> Rebase troubles? stuck_out_tongue

Haha :D yeah i rebased to master :facepalm: TIL about `git reflog`

a8t:
I did the exact same thing the other day but I caught it before force-pushing so I only had to hard reset to origin

fmterrorf:
Nice. I'm so used to merge commits :smile: Trying to get a hold of the rebase workflow

lib/chat_api/conversations.ex
```diff
@@ -496,6 +508,57 @@ defmodule ChatApi.Conversations do
...
   end
 
+  @spec find_or_create_customer_and_conversation(String.t(), String.t()) ::
+          {:ok, Customer.t(), Conversation.t()}
+          | {:error, Ecto.Changeset.t()}
+  def find_or_create_customer_and_conversation(account_id, phone) do
+    now = DateTime.utc_now()
+    sms_source = "sms"
+
+    case Customers.list_customers(account_id, %{phone: phone}) do
```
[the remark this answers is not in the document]
fmterrorf: If if I create a `Customers.find_or_create_by_phone`, It will always return a customer. 

The current code right contains this logic
```
if the customer is not present
    creates a customer + conversation
    return {customer, conversation}
else 
  find conversation
    if no conversation found
        create one
   else
        return {customer, conversation}
```
 If the customer is not present it creates the customer + conversation right away, without having to do a `find` by conversation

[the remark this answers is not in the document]
reichert621: hmm so wouldn't it still work if `Customers.find_or_create_by_phone` just indicated whether the returned customer was `found` or `created`?

maybe something like...

```elixir
case Customers.find_or_create_by_phone(account, phone) do
  {:ok, :found, customer} -> 
    case Conversations.find_latest_conversation_by_customer(customer.id) do
      nil -> Conversations.create_conversation(...)
      conversation -> {:ok, conversation}
    end

  {:ok, :created, customer} -> 
    Conversations.create_conversation(...)

  error -> 
    error
end
```

does that make sense?

reichert621 (approved):
looks good for now! :+1: gonna merge and start playing around with it 😄

lib/chat_api/conversations.ex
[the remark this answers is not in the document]
fmterrorf: Yeah that makes sense!

PR #702, opened 2021-04-02
Fix linking to conversations outside of initial pagination results

reichert621:
### Description

- [x] Fix linking to conversations outside of initial pagination results
- [x] Improve conversations list scroll behavior

### Issue

This broke after we introduced pagination, oops

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/assets/src/components/conversations/ConversationsPreviewList.tsx
+++ b/assets/src/components/conversations/ConversationsPreviewList.tsx
@@ -4,8 +4,9 @@ import {Button, Text} from '../common';
 import ConversationItem from './ConversationItem';
 import ConversationClosing from './ConversationClosing';
 import {getColorByUuid} from './support';
-import {useConversations} from './ConversationsProvider';
 import {Conversation} from '../../types';
+import {isScrolledIntoView} from '../../utils';
+import {useConversations} from './ConversationsProvider';
 
 const ConversationsPreviewList = ({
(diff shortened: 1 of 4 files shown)


assets/src/components/conversations/ConversationsPreviewList.tsx
```diff
@@ -31,6 +32,22 @@ const ConversationsPreviewList = ({
...
+      // Scrolls to highlighted ConversationItem component if not visible
+      const el = document.getElementById(
+        `ConversationItem--${selectedConversationId}`
+      );
+
+      if (!isScrolledIntoView(el)) {
+        el?.scrollIntoView(false);
+      }
+    }
+  }, [selectedConversationId]);
```
reichert621, line 49: note to self: eslint is throwing a warning for adding `conversationIds` as a dependency here, but we don't actually want this to rerun whenever `conversationIds` change... need to figure out the "correct" way to handle this, but otherwise we can just disable the eslint warning 🤷

a8t: you could define `const conversationIdsIncludesSelected = conversationIds.includes(selectedConversationId)` and 

```
React.useEffect(() => {
  if (conversationIdsIncludesSelected) {}
}, conversationIdsIncludesSelected)
```

Issue #591, opened 2021-02-25
Fix UX around team invites


cheeseblubber:
Its confusing to add a team member since people don't read. We should have a plus button under teams that you can input an email and it'll send the invite link to the team member.

rhonsby:
Dibs

PR #703, opened 2021-04-02
Adds ability to send user invitation emails from Account Overview

rhonsby:
### Description

Currently, the only way to invite a user to an account is to generate a an invite url from the Account Overview page, copy that url, and send it directly to the user. We can make this process a lot easier by allowing for account admins to input the user's email and letting us send the invitation url to provided user. 

**Note:** there's a lot of duplicated logic, so I've added some todos. I'm going to address them as a follow-up to this diff, but I wanted to keep them separate to keep the size of this diff smaller.

### Issue

https://github.com/papercups-io/papercups/issues/591

### Screenshots

**Video of invite flow**

https://user-images.githubusercontent.com/1361509/113453069-45997700-93d3-11eb-95d6-c03c56d379bf.mp4

**Email when the sender has a name**
CleanShot 2021-04-02 at 16 32 10@2x

**Email when the sender doesn't have a name**
CleanShot 2021-04-02 at 16 25 19@2x

## Checklist

- [✓] Everything passes when running `mix test`
- [✓] Ran `mix format`
- [✓] No frontend compilation warnings

### DIFF
--- a/lib/workers/send_user_invitation_email.ex
+++ b/lib/workers/send_user_invitation_email.ex
@@ -23,12 +23,24 @@ defmodule ChatApi.Workers.SendUserInvitationEmail do
       account = Accounts.get_account!(account_id)
       Logger.info("Sending user invitation email to #{to_address}")
 
-      ChatApi.Emails.send_user_invitation_email(
-        user,
-        account,
-        to_address,
-        invitation_token
-      )
+      deliver_result =
+        ChatApi.Emails.send_user_invitation_email(
+          user,
+          account,
+          to_address,
+          invitation_token
+        )
+
+      case deliver_result do
+        {:ok, result} ->
+          Logger.info("Successfully sent user invitation email: #{result}")
+
+        {:warning, reason} ->
+          Logger.warn("Warning when sending user invitation email: #{inspect(reason)}")
+
+        {:error, reason} ->
+          Logger.error("Error when sending user invitation email: #{inspect(reason)}")
+      end
     else
       Logger.info("Skipping user invitation email to #{to_address}")
     end
--- a/lib/workers/send_user_invitation_email.ex
+++ b/lib/workers/send_user_invitation_email.ex
@@ -1,0 +1,46 @@
+defmodule ChatApi.Workers.SendUserInvitationEmail do
+  @moduledoc false
+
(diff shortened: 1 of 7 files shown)


lib/workers/send_user_invitation_email.ex
```diff
@@ -0,0 +1,46 @@
...
+    else
+      Logger.info("Skipping user invitation email to #{to_address}")
+    end
+
+    :ok
+  end
+
+  @spec send_user_invitation_email_enabled? :: boolean()
+  def send_user_invitation_email_enabled?() do
+    case System.get_env("USER_INVITATION_EMAIL_ENABLED") do
```
fmterrorf, line 41: nit: should we move this to `config/runtime.exs`, and we could also use `String.to_existing_atom` to cast boolean values. For example: https://github.com/plausible/analytics/blob/master/config/runtime.exs#L68

lib/workers/send_user_invitation_email.ex
```diff
@@ -0,0 +1,46 @@
...
+          "to_address" => to_address,
+          "invitation_token" => invitation_token
+        }
+      }) do
+    if send_user_invitation_email_enabled?() do
+      user = Users.find_by_id!(user_id) |> Repo.preload([:profile])
+      account = Accounts.get_account!(account_id)
+      Logger.info("Sending user invitation email to #{to_address}")
+
+      ChatApi.Emails.send_user_invitation_email(
```
fmterrorf, line 26: `send_user_invitation_email` returns a `deliver_result` that might contain `{:error, binary()}`. Should we log the error when it gets returned?

lib/workers/send_user_invitation_email.ex
rhonsby: I think that's a great idea, but it doesn't seem to be very standard practice in the codebase. We should have a larger discussion about how we should handle environmental variables. FWIW I agree that we should consolidate them into `config/runtime.exs`. 

cc @reichert621

reichert621: yup, let's handle that consolidation in a separate PR :+1:

reichert621 (approved):
woohoooo 🎉 🚀 🔥

Issue #709, opened 2021-04-05
Set up simple form UI for creating a new customer/contact


reichert621:
**Problem**
At the moment, customers are only created when they initialize a conversation (e.g. from chat, email, slack etc)

We want to be able to create customers from scratch, so we can do things like add notes, tags, proactively send them messages, etc.

**Solution**
Set up a simple form UI that creates a new customer from the following fields:
- email
- name (optional)
- phone (optional)
- external_id (optional)

We can probably just do this in a modal on the `CustomersPage`... not too picky about the UI!

**Testing**
Please include screenshots/gifs in your PR :) 

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

PR #717, opened 2021-04-06
Set up modal to create new customer

reichert621:
### Description

This PR sets up a simple modal for creating new customers.

### Issue

https://github.com/papercups-io/papercups/issues/709

### Screenshots

Screen Shot 2021-04-06 at 2 22 01 PM

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/assets/src/components/customers/CustomersPage.tsx
+++ b/assets/src/components/customers/CustomersPage.tsx
@@ -2,12 +2,14 @@ import React from 'react';
 import {Link} from 'react-router-dom';
 import {Box, Flex} from 'theme-ui';
 import {Alert, Button, Input, Paragraph, Text, Title} from '../common';
+import {PlusOutlined} from '../icons';
 import {useConversations} from '../conversations/ConversationsProvider';
 import * as API from '../../api';
-import Spinner from '../Spinner';
-import CustomersTable from './CustomersTable';
 import logger from '../../logger';
 import {Customer} from '../../types';
+import Spinner from '../Spinner';
+import CustomersTable from './CustomersTable';
+import NewCustomerModal from './NewCustomerModal';
 
 const filterCustomersByQuery = (
   customers: Array<Customer>,
@@ -49,6 +51,7 @@ type Props = {
 type State = {
   loading: boolean;
   refreshing: boolean;
+  isModalOpen: boolean;
   selectedCustomerId: string | null;
   query: string;
   customers: Array<Customer>;
@@ -59,6 +62,7 @@ class CustomersPage extends React.Component<Props, State> {
   state: State = {
     loading: true,
     refreshing: false,
+    isModalOpen: false,
     selectedCustomerId: null,
     query: '',
     customers: [],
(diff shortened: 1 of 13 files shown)


assets/src/components/customers/CustomersTable.tsx
```diff
@@ -101,19 +104,11 @@ const CustomersTable = ({
...
-        const {browser, os} = record;
-
-        return (
-          <Text>
-            <Text type="secondary">{browser}</Text>
-            {browser && os ? ' · ' : ''}
-            {os && <Text type="secondary">{os}</Text>}
-          </Text>
-        );
+      title: 'Timezone',
```
reichert621, line 107: (figured this information is more useful 🤷 )

assets/src/components/customers/CustomersPage.tsx
```diff
@@ -156,14 +179,28 @@ class CustomersPage extends React.Component<Props, State> {
...
+            >
+              New customer
+            </Button>
+          </Flex>
+
+          <NewCustomerModal
+            visible={isModalOpen}
+            onCancel={this.handleCreateCustomerCancelled}
+            onSuccess={this.handleCreateCustomerSuccess}
+          />
```
rhonsby, line 203: Nit: probably not necessary for this diff, but it might be worthwhile to consider consolidating button + modal logic into a separate component so that the current component (i.e. CustomersPage) doesn't have to also hold all of this logic related to whether the modal is open (see https://github.com/papercups-io/papercups/pull/718).

assets/src/components/customers/NewCustomerModal.tsx
```diff
@@ -0,0 +1,126 @@
...
+        resetInputFields();
+      })
+      .catch((err) => {
+        logger.error('Error creating customer:', err);
+        // TODO
+        const errorMessage = 'Invalid customer details!';
+        setErrorMessage(errorMessage);
+      })
+      .finally(() => setIsSaving(false));
+  };
```
rhonsby: Is there a reason why we're handling the promise traditionally with then/catch/finally vs awaiting the promise and wrapping the logic in a try/catch?

assets/src/types.ts
```diff
@@ -32,11 +32,14 @@ export type Customer = {
   first_seen?: any;
   host?: string;
   ip?: string;
+  // Date
   last_seen: string;
+  // Datetime
+  last_seen_at?: string;
```
rhonsby, line 38: Just curious - is there a reason why we have two `last_seen` properties? From what I can tell, they're different formats (date/datetime), but is there a reason why we don't just always have `last_seen_at` and make sure that it's not optional? I imagine that it would be easier if there was only one property.

reichert621: yeah i'd like to deprecate the `last_seen` date property, it was something we set up really early on and haven't had a chance to remove it in favor of `last_seen_at`

assets/src/components/customers/CustomersPage.tsx
reichert621: good idea :+1:

assets/src/components/customers/NewCustomerModal.tsx
reichert621: no reason really, i find both approaches interchangeable in many cases 🤷

rhonsby:
PR looks good to me :) Let's make sure we have a follow-up issue for deprecating `last_seen` properly since that can bite us later on if we don't clean that up

assets/src/components/customers/CustomersPage.tsx
reichert621: done!

Issue #688, opened 2021-03-25
Proactively send messages to customers from the customer details page


reichert621:
**Problem**
Users want to be able to proactively send messages to their customers.

Now that we have a customer details page, let's add this feature there!

**Solution**
Add a modal (or some UI) that enables sending a message to a customer that hasn't started a conversation yet.

**Additional context**
At the moment this feature only exists within the Live Session viewer (with our Storytime feature).

**Testing**
This will require a bit of manual testing.

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

taewookim:
might be better to just use setTimeout() and pop a chat ANYWAY  with a pre-written message in case of user idling, so that it engages the user

(FYI: im the one who requested this feature on slack chat with Alex)

taewookim:
How do you enable this feature? im using the hosted version of papercups

reichert621:
@taewookim we'll be deploying some updates today :) but here's the general flow:
![proactive-chat-demo](https://user-images.githubusercontent.com/5264279/116591764-3a137000-a8ed-11eb-83c0-4efcb4e5d331.gif)

PR #718, opened 2021-04-06
Adds button to start a new conversation from the customer details page

rhonsby:
### Description
The high level goal for this pull request is to add the ability for users to proactively send messages to customers from the customer details page. 

In order to do that, I refactored the button and modal that existed to start new conversations from the live session viewer page into a separate component (4eacc9b). With just the customer id, we're not able to start new conversations anywhere in the app.

4b5537a uses the new component to add the ability to start a new conversation on the customer details page.

Notice that on the live session viewer page, I also made the design decision to always show the button, but disable it with a tooltip when a conversation is already in progress. This teaches our users to reference existing conversations before creating new ones. Hiding buttons based on certain states isn't always clear for our users and can lead to a confusing experience.

### Issue
https://github.com/papercups-io/papercups/issues/688

### Screenshots
**Starting a new conversation from the live session page**
https://user-images.githubusercontent.com/1361509/113760331-168d4900-96e4-11eb-9041-c80e01aa61db.mp4

**Starting a new conversation from the customer details page**
https://user-images.githubusercontent.com/1361509/113760340-18efa300-96e4-11eb-8fec-d2a0b5cbaf3a.mp4

## Checklist

- [✔] Everything passes when running `mix test`
- [✔] Ran `mix format`
- [✔] No frontend compilation warnings

### DIFF
--- a/assets/src/components/conversations/StartConversationButton.tsx
+++ b/assets/src/components/conversations/StartConversationButton.tsx
@@ -1,0 +1,139 @@
+import React from 'react';
+
+import {Button, Modal, TextArea, Tooltip} from '../common';
+import {SendOutlined} from '../icons';
+import * as API from '../../api';
+import logger from '../../logger';
+import {Conversation} from '../../types';
+import {TooltipPlacement} from 'antd/lib/tooltip';
+
+export type Props = {
+  customerId: string;
+  disabled: boolean;
+  disabledTooltipPlacement: TooltipPlacement;
+  disabledTooltipTitle: string;
+  onInitializeNewConversation?: (conservation: Conversation) => void;
+};
+
+export type State = {
+  isConversationModalOpen: boolean;
+  isSendingMessage: boolean;
+  message: string;
+};
+
+export class StartConversationButton extends React.Component<Props, State> {
+  static defaultProps = {
+    disabled: false,
+    disabledTooltipPlacement: 'right',
+    disabledTooltipTitle: 'This customer already has an open conversation',
+  };
+
+  state: State = {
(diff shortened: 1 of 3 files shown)


assets/src/components/conversations/StartConversationButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+        onInitializeNewConversation(conversation);
+      }
+    } catch (err) {
+      logger.error('Failed to initialize conversation!', err);
+    }
+
+    this.setState({isConversationModalOpen: false, isSendingMessage: false});
+  };
+
+  getButton = () => {
```
reichert621, line 69: nit - can we just put this in a separate component?

assets/src/components/conversations/StartConversationButton.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const {
+      disabled,
```
reichert621, line 71: very minor nitpick - can we call this `isDisabled` instead so it doesn't conflict with the html "disabled" attribute (https://www.w3schools.com/tags/att_disabled.asp)

(i think 99% of the time this doesn't matter, but i remember at stripe we had a couple issues with calling a prop `disabled` and had a rule that we would preferably use `isDisabled` instead)

reichert621 (commented):
nice! very excited about this 😄 

just a couple nitpicks!

reichert621:
(just a heads up @rhonsby, i rebased with master and i'm deploying to staging to test it out -- hoping to do a prod deploy today and have this included!)

PR #732, opened 2021-04-12
Adds initial implementation of customer details page redesign

rhonsby:
### Description

This commit consists of the initial implementation of the customer details page redesign. 

Note that I hid the redesign behind an env variable because it isn't complete in its implementation. I'm doing this for the purpose of ensuring that these commits are smaller and more reviewable. There's still a few missing features:
* Ability to edit basic properties like name, email, phone number, etc.
* Ability to delete customer
* Ability to export data
* Ability to view notes

#### Before
image

#### After

image

#### After (non-live customer and unattached to company)
image

## Checklist

- [x] Ran `mix format`


assets/src/api.ts
```diff
@@ -1269,7 +1270,7 @@ type BrowserSessionFilters = {
 export const fetchBrowserSessions = async (
   {customerId, isActive, limit = 100, sessionIds = []}: BrowserSessionFilters,
   token = getAccessToken()
-) => {
+): Promise<Array<BrowserSession>> => {
```
rhonsby, line 1273: We should ideally type all of our API responses, right?

assets/src/components/Dashboard.tsx
```diff
@@ -421,7 +423,14 @@ const Dashboard = (props: RouteComponentProps) => {
             component={GettingStartedOverview}
           />
           <Route path="/account*" component={AccountOverview} />
-          <Route path="/customers/:id" component={CustomerDetailsPage} />
+          <Route
+            path="/customers/:id"
+            component={
+              isCustomerDetailsV2Enabled
+                ? CustomerDetailsPageV2
+                : CustomerDetailsPage
```
rhonsby, line 431: As mentioned in the description, I'm hiding the redesign behind an env variable while I work on implementing the rest of the features. I didn't want to have these changes sitting on a branch for too long.

reichert621: perfect! was going to recommend this approach anyway :+1:

reichert621: we could also just create a new temporary route for v2, like `/v2/customers/:id` -- might make it easier to view both 🤷

assets/src/components/customers/CustomerUtils.ts
```diff
@@ -0,0 +1,15 @@
+import * as API from '../../api';
+import logger from '../../logger';
+import {download} from '../../utils';
+
+export const exportCustomerData = async (customerId: string) => {
```
rhonsby, line 5: Seemed like a fairly component-agnostic utility function so I extracted it into its own utils file.

lib/chat_api_web/views/customer_view.ex
```diff
@@ -50,7 +50,8 @@ defmodule ChatApiWeb.CustomerView do
       current_url: customer.current_url,
       browser: customer.browser,
       os: customer.os,
-      metadata: customer.metadata
+      metadata: customer.metadata,
+      title: customer.name || customer.email || "Anonymous User"
```
rhonsby, line 54: I noticed we had similar logic in different places, but would render different content. I figured this would be a good way to consolidate that so we don't have to duplicate that logic in too many places.

assets/src/components/Dashboard.tsx
rhonsby: Yeah, that sounds like a good idea! I'll add that. It'll be nice to be able to view it in prod.

assets/src/components/customers/CustomerDetailsSidebar.tsx
```diff
@@ -0,0 +1,316 @@
...
+      <Box mb={2}>
+        <Text strong>{title}</Text>
+      </Box>
+      {children}
+    </Box>
+  );
+};
+
+const CustomerDetailsProperty = ({
+  Icon,
```
reichert621, line 242: nit -- i think react props should always be lowercase? 🤔 

but i see this is passing in the icon class/function rather than the component itself, so i see why it's capitalized... i think i'd rather pass in the component itself even if we need to explicitly set the color every time, e.g.
```jsx
<CustomerDetailsProperty
  icon={<LinkOutlined style={{color: colors.primary}} />}
  name="Website"
  value={websiteUrl}
/>
```

assets/src/components/customers/CustomerDetailsSidebar.tsx
```diff
@@ -0,0 +1,316 @@
...
+  name,
+  value,
+}: {
+  Icon: React.ComponentType<any>;
+  name: string;
+  value: any;
+}) => {
+  let valueComponent;
```
reichert621, line 250: hmm i wonder if similar to the `icon` prop we should just enforce that this is a `ReactElement` type and explicitly pass in `<Text>{value || 'Unknown'}</Text>` when we need to 🤔 i kinda prefer the explicit nature of it, and in some cases we might want the fallback value to be something other than "unknown" 

(this behavior might also be confusing if a value of `0` is passed in, for example... though i suppose this edge case could be fixed by using `value ?? 'Unknown'` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator)

otherwise, small nitpick but we can probably just assign the `valueComponent` in a ternary expression, e.g.
```js
const valueComponent = React.isValidElement(value) ? value : <Text>{value || 'Unknown'}</Text>;
```

assets/src/components/customers/CustomerDetailsSidebar.tsx
```diff
@@ -0,0 +1,316 @@
...
+      {valueComponent}
+    </Flex>
+  );
+};
+
+const getLastSeenValue = ({
+  isOnline,
+  lastSeenAt,
+}: {
+  isOnline: boolean | undefined;
```
reichert621, line 282: i _think_ in typescript it's a bit more idiomatic to use `isOnline?: boolean` instead of `isOnline: boolean | undefined` (i'm pretty sure they're the same thing, but i might be wrong)

assets/src/components/customers/CustomerUtils.ts
```diff
@@ -0,0 +1,15 @@
+import * as API from '../../api';
```
reichert621, line 1: nit -- for the file name, we've been using the convention of calling util files like this `customers/support.ts` (just a convention we used at stripe which i've kept)

assets/src/components/customers/CustomersTable.tsx
```diff
@@ -4,7 +4,7 @@ import dayjs from 'dayjs';
 import utc from 'dayjs/plugin/utc';
 import {Customer} from '../../types';
 import {Badge, Button, Table, Text, Tooltip} from '../common';
-import CustomerDetailsModal from './CustomerDetailsModal';
+import CustomerDetailsModal from '../customers/CustomerDetailsModal';
```
reichert621, line 7: was this change necessary? 🤔

reichert621: (on an unrelated note, it might be interesting to switch to using absolute(?) paths rather than relative paths)

lib/chat_api_web/views/customer_view.ex
reichert621: seems reasonable!

reichert621 (changes requested):
looks great @rhonsby! very excited about these design improvements 🔥 

just some nitpicky stuff, curious to hear your thoughts :)

assets/src/components/customers/CustomerDetailsCard.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import {Box} from 'theme-ui';
+import {SxStyleProp} from 'theme-ui';
```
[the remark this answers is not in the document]
rhonsby: Yeah, absolutely! Good catch.

assets/src/components/customers/CustomerDetailsSidebar.tsx
```diff
@@ -0,0 +1,316 @@
...
+        <CustomerDetailsSection title="Tags">
+          <SidebarCustomerTags customerId={customerId} />
+        </CustomerDetailsSection>
+      </Box>
+    </CustomerDetailsCard>
+  );
+};
+
+const CompanyDetailsSection = ({company}: {company: Company | undefined}) => {
+  let body;
```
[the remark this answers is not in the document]
rhonsby: Yeah, sure thing :)

assets/src/components/customers/CustomerDetailsSidebar.tsx
rhonsby: It's been a hot minute so I forgot that that was a best practice 😅 It didn't seem to be a problem, but I'm happy to change it to lowercase. 

Thoughts on a helper method used to reduce duplication? 

```
const getIconForComponent = (Icon) => (<Icon style={{color: color.primary}}>);
```

assets/src/components/customers/CustomerDetailsSidebar.tsx
rhonsby: I considered that as well. I had it set up this way to reduce code duplication, but I'm not actually a huge fan of code paths where different types render different results (harder to reason about). Let's err on the side of being explicit. I'll change it 👌

assets/src/components/customers/CustomerDetailsSidebar.tsx
rhonsby: Yeah, you're right. In this case, it isn't a big deal so I can change it to shorthand it, but there's one case where it actually does make a difference.

```
const firstFn = ({a, b}: {a: string, b: string | undefined}) => {
    console.log({a, b});
};

const secondFn = ({a, b}: {a: string, b?: string}) => {
    console.log({a, b});
};
```

In `firstFn`, `b` is required, but can be a `string` or `undefined`. In `secondFn`, `b` isn't required, but if it's there, it can be either `string` or `undefined`. 

How often does it happen where a property is required, but can be `undefined`? I'm not sure, but it's good to know the distinction. Again, for this case, it doesn't matter 🙂

assets/src/components/customers/CustomerUtils.ts
rhonsby: Ooo ok. I dig it. Will update!

assets/src/components/customers/CustomersTable.tsx
rhonsby: No, I'm not sure why this changed. I must have moved some stuff around and accidentally made this change. I'll fix it. 

re: absolute paths - I don't have an opinion either way.

assets/src/components/customers/CustomerDetailsSidebar.tsx
reichert621: down for that! :+1:

assets/src/components/customers/CustomerDetailsSidebar.tsx
reichert621: ahh good point, i'd say if a property can be undefined than it's nice to have it not be required :)

reichert621 (approved):
nice! 🎉

Issue #720, opened 2021-04-06
Deprecate customer `last_seen` field


reichert621:
**Problem**
This field is no longer helpful, now that we have `last_seen_at` as a datetime (rather than `last_seen` which is just a date)

**Solution**
- Find where this field is used in the code and get rid of it or replace it with `last_seen_at` where possible
- Write the db migration to remove the field

**Additional context**
This is currently a required field on the `customers` table, but I'm not sure why...

**Testing**
Make sure tests pass, and try to write some additional tests where necessary!

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

reichert621:
(cc @rhonsby)

utsav0209:
Hey @reichert621, I am planning to work on this if no one else is already working on this.

reichert621:
sure @utsav0209! thanks for putting up the PR :) assigning you to this ticket now :+1:

PR #734, opened 2021-04-13
Remove last_seen field from customer

utsav0209:
### Description

Dropped required field last_seen from Customer table and added last_seen_at as required field

### Issue

#720 

## Checklist

- [x] Ran `mix format`


assets/src/components/customers/CustomersTable.tsx
```diff
@@ -77,8 +77,10 @@ const CustomersTable = ({
       dataIndex: 'last_seen_at',
       key: 'last_seen_at',
       render: (value: string, record: Customer) => {
-        const {id, pathname, current_url, last_seen} = record;
-        const formatted = dayjs.utc(value || last_seen).format('MMMM DD, YYYY');
+        const {id, pathname, current_url, last_seen_at} = record;
+        const formatted = dayjs
+          .utc(value || last_seen_at)
```
reichert621, line 82: `value` and `last_seen_at` should be the same in this case, so we can just get rid of the `|| last_seen_at` here i believe

lib/chat_api/customers/customer.ex
```diff
@@ -13,7 +13,7 @@ defmodule ChatApi.Customers.Customer do
 
   @type t :: %__MODULE__{
           first_seen: any(),
-          last_seen: any(),
+          last_seen_at: any(),
```
reichert621, line 16: i think this is already set below on line 28, so we can just remove this line

reichert621 (changes requested):
thanks for handling this 🎉 just a couple small things but otherwise looks good!

priv/repo/migrations/20210414003100_drop_last_seen_customer.exs
```diff
@@ -0,0 +1,9 @@
+defmodule ChatApi.Repo.Migrations.DropCusotmerLastSeen do
+  use Ecto.Migration
+
+  def change do
+    alter table(:customers) do
+      remove :last_seen
```
reichert621, line 6: we should probably also ensure that any records with a missing `last_seen_at` value get populated with whatever was in `last_seen`

utsav0209: Oops, I took a pull from master and realized you have already done this!

utsav0209:
@reichert621 any comments on this?

priv/repo/migrations/20210417003100_drop_last_seen_customer.exs
```diff
@@ -0,0 +1,12 @@
+defmodule ChatApi.Repo.Migrations.DropCusotmerLastSeen do
+  use Ecto.Migration
+  import Ecto.Query
+
+  alias ChatApi.Repo
+
```
reichert621, line 6: @utsav0209 mind fixing the warnings above?

reichert621: i'm also thinking maybe we shouldn't drop this column quite yet -- maybe we can handle this migration in a separate PR to give us time to test that everything is working as expected?

reichert621 (commented):
@utsav0209 just a couple small comments!

i'll also want to do some manual testing on this to make sure it doesn't break anything before we deploy to production 👍

priv/repo/migrations/20210417003100_drop_last_seen_customer.exs
utsav0209: Yeah, I think that's a good idea, can put PR later for this migration once we are sure things are working as expected. I will remove the migration from this PR in that case

utsav0209: @reichert621 I am also removing the required check for the last_seen field for this to move forward

utsav0209:
@reichert621 any updates on this?

reichert621 (approved):

reichert621:
thanks for the help on this @utsav0209! ❤️  (sorry for the delay)

Issue #750, opened 2021-04-19
Make customer notes section on customer page bigger.


cheeseblubber:
It would also be nice if the notes on the customer details page...
- displayed the author/creator of the note somewhere
- displayed the timestamp of the note (when it was last updated?) somewhere
- allowed for line breaks (instead of auto-saving on "enter")
- (optional) support markdown? (see example of `ReactMarkdown` used here: https://github.com/papercups-io/papercups/blob/master/assets/src/components/conversations/ChatMessageBox.tsx#L2)

(cc @rhonsby)

PR #763, opened 2021-04-20
Adds markdown rendering to customer notes

rhonsby:
### Description

Similarly to chat messages, we want to be able to write and render markdown in customer notes. 

Because we're introducing markdown, we need to be able to click on things like links. Instead of trigger the delete confirmation when the whole note is clicked, I introduced a Delete link in the top right corner.

### Issue

https://github.com/papercups-io/papercups/issues/750

### Screenshots

image

## Checklist

- [x] Ran `mix format`

### DIFF
--- a/assets/src/components/customers/CustomerDetailsNotes.tsx
+++ b/assets/src/components/customers/CustomerDetailsNotes.tsx
@@ -3,7 +3,6 @@ import dayjs from 'dayjs';
 import {Box, Flex} from 'theme-ui';
 
 import {
-  Button,
   colors,
   Divider,
   Empty,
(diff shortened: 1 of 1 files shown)


assets/src/components/customers/CustomerDetailsNotes.tsx
```diff
@@ -2,7 +2,15 @@ import React from 'react';
 import dayjs from 'dayjs';
 import {Box, Flex} from 'theme-ui';
 
-import {colors, Divider, Empty, Popconfirm, Text} from '../common';
+import {
+  Button,
```
reichert621, line 6: (looks like this is unused, which is breaking the linter)

reichert621 (changes requested):
just an unused import breaking `npm run build`

assets/src/components/customers/CustomerDetailsNotes.tsx
reichert621, line 6:
```suggestion
```

Issue #760, opened 2021-04-20
Introduce a concept of inboxes


rhonsby:
**Problem**
As we add support for multiple channels like SMS, support emails, shared Slack channels, live chats, etc., we need a way to make it easier for users to view conversations based on the channels to better manage them. 

For example, live chats are a bit more time sensitive than channels like SMS, emails, etc. By separating the inboxes and making it easy to see which ones have unread messages, users can more easily manage which ones they reply to. 

In the future, we'll also want users to be able to create multiple inboxes that use the same channels. For example, a user might have two chat widgets, one on the landing page and one in their app, which should direct messages to two different live chat inboxes. 

**Solution**
Our goal is to have a separate inbox per channel. For the v0 implementation, I want to do the simplest thing possible:

### Frontend
* Create a submenu in the navigation that with menu items for each channel (e.g. email, slack, chat). Instead of `/conversations/all`, we'll follow a format of `/conversations/<source>` (e.g. `/conversations/slack`, `/conversations/email`, `/conversations/chat`). 
* Re-use the `ConversationDashboard` component used for `/conversations/all`, extend it to accept the `source`, and pass the `source` value to the server when querying for conversations.

### Backend
* Update `conversation_controller.ex#index` to make use of the `source` to filter the results.

**Describe alternatives you've considered**
No other alternatives considered, but open to suggestions.

**Testing**
This will be tested by creating multiple conversations with varying sources, navigating to each inbox, and verifying that the conversations are rendered in the correct inboxes.

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

reichert621:
@rhonsby this seems great! i'm assuming that once conversations are closed from these new inboxes, they just go to the "Closed" section of the primary inbox?

seems like a good first step :+1: eventually we'll need a concept of an `inbox` model but for now having these "filters by channel" will be helpful for us

rhonsby:
@reichert621 

> once conversations are closed from these new inboxes, they just go to the "Closed" section of the primary inbox?

Yeah, that's basically what would happen since the "inboxes" right now are really just filters on the source. 

> eventually we'll need a concept of an inbox model

[...]

PR #788, opened 2021-04-26
Refactors how conversations are stored in ConversationsProvider.tsx

rhonsby:
### Description

This is part 2 of refactoring `ConversationsProvider.tsx`. This commit specifically refactors how conversations are stored at the top level into an `inboxes` bucket. This will make it easier for us to continue to adding new inboxes.

### Issue

https://github.com/papercups-io/papercups/issues/760

### Screenshots

https://user-images.githubusercontent.com/1361509/116147217-722c7000-a6ad-11eb-8d54-affa209f3255.mp4

## Checklist

- [ ] Ran `mix format`


assets/src/components/conversations/ConversationsProvider.tsx
```diff
@@ -16,20 +16,55 @@ import {
 } from '../../presence';
 import ConversationNotificationManager from './ConversationNotificationManager';
 
+type Inbox = {
+  conversationIds: string[];
+  unreadCount: number;
+};
+
+type Inboxes = {
```
rhonsby, line 24: This structure will make it easier for us to expand to adding custom inboxes as well. 

* `all` represents all of the inboxes we support today.
* `bySource` will contain all of the inboxes that are channel specific.
* We can expand by adding something like `custom` which will contain all of the inboxes that the user configured themselves.

assets/src/components/conversations/ConversationsProvider.tsx
```diff
@@ -374,19 +393,8 @@ export class ConversationsProvider extends React.Component<Props, State> {
...
-        const left = a.last_activity_at
-          ? +new Date(a.last_activity_at)
-          : -Infinity;
-        const right = b.last_activity_at
-          ? +new Date(b.last_activity_at)
-          : -Infinity;
-
-        return right - left;
-      }
+    const sortedConversationIds = this.getSortedConversationIds(
```
rhonsby, line 396: Note that I migrated the logic for sorting conversations into `getSortedConversationIds`.

assets/src/components/conversations/ConversationsProvider.tsx
```diff
@@ -417,47 +421,11 @@ export class ConversationsProvider extends React.Component<Props, State> {
...
-    const mine = openConversations
-      .filter((c) => c.assignee_id === currentUserId && c.status === 'open')
-      .map((c) => c.id);
-    const priority = openConversations
-      .filter((c) => c.priority === 'priority' && c.status === 'open')
-      .map((c) => c.id);
-    const closed = allConversations
-      .filter((c) => c.status !== 'open')
-      .map((c) => c.id);
+    const inboxes = this.getInboxes(updatedConversationsById);
```
rhonsby, line 424: I've noticed that our codebase has a few pretty big functions here and there. I started to clean this up by creating a lot of smaller functions. Let me know what your thoughts are on many small functions vs fewer larger functions.

reichert621: yup i prefer this as well :+1:

reichert621 (changes requested):
just a couple small things i noticed while testing the branch!

assets/src/components/Dashboard.tsx
```diff
@@ -441,6 +496,15 @@ const Dashboard = (props: RouteComponentProps) => {
             path="/conversations/priority"
             component={PriorityConversations}
           />
+          <Route path="/conversations/live-chat">
+            <ConversationsBySource title="Live chat" source="chat" />
+          </Route>
+          <Route path="/conversations/email">
+            <ConversationsBySource title="Email" source="email" />
```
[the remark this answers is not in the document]
rhonsby: It looks like adding `key` to the `Route` fixed the issue. Can you make sure that it works for you too?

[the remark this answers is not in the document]
reichert621: seems to work for me too! :+1:

reichert621 (approved):

Issue #774, opened 2021-04-22
Dashboard websocket disconnects and doesn't load new messages after a long period


cheeseblubber:
**Describe the bug**
Dashboard doesn't load new messages after a certain amount of time. If you send a message in to the account it doesn't load until you refresh

In the JS console it loads websocket errors
<img width="537" alt="Screen Shot 2021-04-22 at 7 25 04 PM" src="https://user-images.githubusercontent.com/4218509/115797078-0a221500-a3a1-11eb-9357-d2dd17d71755.png">

**To Reproduce**
1. Load dashboard either app.papercups.io or localhost 
2. Put your computer to sleep and revisit after x minutes (Not sure the actual minutes I would start with 15) I've seen it definitely happen after an hour
3. See console error

**Expected behavior**
The dashboard should either refresh the websockets or prompt the user to refresh the page. 

Ideally it should refresh the websockets but if it is too difficult prompting for a refresh is a good temporary work around

**Additional context**
The way I've reproduced it is by closing my Mac and keeping the dashboard open. I've reproduced it after an hour being away during lunch. It should disconnect with less time.

PR #792, opened 2021-04-27
Adds better error reporting when socket errors and try to reconnect

rhonsby:
### Description

When the socket used for managing conversations errors out, we currently only log an error message. This commit logs the error as well to provide more visibility into why it's erroring and attempts to reconnect when it does error.

### Issue

https://github.com/papercups-io/papercups/issues/774

## Checklist

- [ ] Ran `mix format`

### DIFF
--- a/assets/src/components/conversations/ConversationNotificationManager.ts
+++ b/assets/src/components/conversations/ConversationNotificationManager.ts
@@ -24,6 +24,7 @@ class ConversationNotificationManager {
   config: Config;
   socket: Socket;
   channel: Channel | null = null;
+  reconnectIntervalId?: number;
 
   constructor(config: Config) {
     this.socket = this.createNewSocket();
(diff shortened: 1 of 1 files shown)


reichert621:
@rhonsby just curious, were you able to see if the reconnecting logic fixes anything when the socket gets disconnected (e.g. after the computer goes to sleep and gets reopened)?

rhonsby:
@reichert621 I couldn't get the socket to disconnect with a short-ish sleep, so I'm going to try to let it sleep over lunch. I tested it another way that I thought would be similar:

1. Load the page and verify in the console that you're connected. 
2. Try sending a message and verify that it works.
2. Force stop your server and observe that the socket error handler is triggered.
3. Try sending a message and verify that it doesn't work.
4. Restart your server, but make sure you click the stop button on the browser to keep the page from refreshing.
5. Observe that you reconnect back (via console messages and trying to send a message).

Issue #822, opened 2021-05-06
Starting conversations programatically


lucasavila00:
**Problem**
I want to be able to send an user a message when my server decides it. It shouldn't matter if the user has started a conversation yet or not.

**Solution**
There are two approaches:
- Server to Server
- Client Based

A server to server solution would be the best one, but it would require a lot of work on Papercups to support it. As it is, the client won't register itself until a message has been sent.

It seems a client-based solution, as it would require just a special type of user message to flow through the same paths, however this would be unsafe for sensitive data, since at the end my server message will flow through the client machine before reaching Papercups. These messages could be signed in my server beforehand, making it safe.

The client-side approach seems the best one in relation to not having to sync what my system has as user_id to a unique Papercups customer_id.

In the end, the client-side approach would expose a JS API like

`
Papercups.sendMessageAsServer("blablabla");
Papercups.sendSignedMessageAsServer("blablabla");
`

which is my responsibility to call at the right time. This message would then be sent to the chat as usual and be shown to the person answering and the user in the chat.

As I see it, the only downside of the client-based approach is that Papercups can't be used to reach the user if the conversation is closed, but it also doesn't require an unique user to be set up in agreement between Papercups and my server.

lucasavila00:
I have made this draft of what are the needed changes, and they're not that much.

The signature handling part on Elixir was left vague as many technologies could work there.

PR #823, opened 2021-05-07
[Draft] Client Side Bot Messages

lucasavila00:
### Issue

https://github.com/papercups-io/papercups/issues/822

### Screenshots

## Checklist

- [ ] Ran `mix format`

### DIFF
--- a/assets/src/components/billing/BillingOverview.tsx
+++ b/assets/src/components/billing/BillingOverview.tsx
@@ -37,7 +37,7 @@ import {LITE_PRICE, STARTER_PRICE, TEAM_PRICE} from '../../constants';
 import {env} from '../../config';
 import './Billing.css';
 
-const stripe = loadStripe(env.REACT_APP_STRIPE_PUBLIC_KEY);
+const stripe = loadStripe('undefined');
 
 const BillingBreakdownTable = ({
   loading,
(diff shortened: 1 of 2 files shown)


assets/src/components/billing/BillingOverview.tsx
```diff
@@ -37,7 +37,7 @@ import {LITE_PRICE, STARTER_PRICE, TEAM_PRICE} from '../../constants';
 import {env} from '../../config';
 import './Billing.css';
 
-const stripe = loadStripe(env.REACT_APP_STRIPE_PUBLIC_KEY);
+const stripe = loadStripe('undefined');
```
lucasavila00, line 40: this was crashing the app

it should be fixed before removing the draft tag

lib/chat_api_web/channels/conversation_channel.ex
```diff
@@ -161,4 +190,8 @@ defmodule ChatApiWeb.ConversationChannel do
       _ -> false
     end
   end
+
+  defp user_id_of_signature(_signature) do
+    1
```
lucasavila00, line 195: it should receive the message, signature and validate it

if valid, return an user_id so that the message looks like it's from an agent?

Issue #829, opened 2021-05-10
Email support form 


cheeseblubber:
**Problem**
Sometimes our users want an email support form instead of a chat widget and I personally hate mailto and have never used it. We want to make it really easy for developers to set up an email form and customize it.

**Solution**
1. Make a support form inside of the chatwidget that sends an email to a email address that you specify in Papercups. 
2. Make it easy to set up with a reactt/vue/html components.  Start off with react and html
3. Specify the email address the email form goes to

PR #835, opened 2021-05-12
Adds endpoint for starting a new email conversation

rhonsby:
### Description

Commit 1/N for supporting the ability to send an email via a form. This commit adds the API endpoint. Will follow up with a commit to send an email receipt to the user.

### Issue

https://github.com/papercups-io/papercups/issues/829

## Checklist

- [x] Ran `mix format`


lib/chat_api/customers.ex
```diff
@@ -246,7 +246,7 @@ defmodule ChatApi.Customers do
   @spec create_customer(map()) :: {:ok, Customer.t()} | {:error, Ecto.Changeset.t()}
   def create_customer(attrs \\ %{}) do
     %Customer{}
-    |> Customer.changeset(attrs)
+    |> Customer.changeset(Map.merge(get_default_params(), attrs))
```
rhonsby, line 249: I noticed that we weren't setting default props, but we do it for a lot of the other create methods. lmk if you think this shouldn't be here.

lib/chat_api_web/controllers/email_conversation_controller.ex
```diff
@@ -0,0 +1,91 @@
...
+        "account_id" => account_id,
+        "body" => body,
+        "customer_id" => customer_id,
+        "email_address" => email_address,
+        "subject" => subject
+      })
+    end
+  end
+
+  def create(
```
rhonsby, line 35: I'm still not quite used to the idea of pattern matching, but I'm using it here to pattern match on the existence of `customer_id` in the params. If it exists, then we assume that the customer exists in the database and we update it with the provided email address. If it doesn't, then we create the customer.

rhonsby:
@reichert621 here's a wip so far. lmk if you have any questions thoughts.

lib/chat_api/customers.ex
```diff
@@ -196,15 +196,18 @@ defmodule ChatApi.Customers do
...
 
-  @spec find_by_email(binary() | nil, binary()) :: Customer.t() | nil
-  def find_by_email(nil, _account_id), do: nil
+  @spec find_by_email(binary() | nil, binary(), atom() | list(atom()) | keyword()) ::
+          Customer.t() | nil
+  def find_by_email(email, account_id, preloads \\ [])
+  def find_by_email(nil, _account_id, _preloads), do: nil
 
-  def find_by_email(email, account_id) do
+  def find_by_email(email, account_id, preloads) do
```
rhonsby, line 204: Used in this commit to preload messages in tests, but I imagine that this could be useful in normal application code as well.

lib/chat_api_web/controllers/email_conversation_controller.ex
```diff
@@ -0,0 +1,92 @@
...
+        %{
+          "account_id" => account_id,
+          "body" => body,
+          "customer_id" => customer_id,
+          "email_address" => email_address
+        } = params
+      ) do
+    subject = Map.get(params, "subject")
+
+    with customer <- Customers.get_customer!(customer_id),
```
rhonsby, line 21: What's the best practice here for handling a `customer_id` id that doesn't exist? The database throws an exception if this happens, but I didn't see very much error handling in the codebase for this.

rhonsby:
@reichert621 Sorry for totally leaving this hanging. Do you want to take this work over while I'm out? I don't want you guys to wait on me if you want this done before I get back.

reichert621:
No worries! It’s not super urgent, but if it can’t wait I’m happy to pick
it up :) don’t worry about it though

Issue #769, opened 2021-04-22
Support SMTP for transactional emails 


cheeseblubber:
**Problem**
Currently for self hosting some users don't want to use mailgun or a third party service for sending transactional emails. 

**Solution**
We should support SMTP as an integration so the self hosted users can provide their own service. The scope of this feature is purely for sending emails and not receiving.

We should also set the email hosted service based on the environment variable that is being set. 
For example if you set the environment variables: 

SMTP_SERVER
SMTP_HOSTNAME
SMTP_PORT etc.

It should use SMTP as your transactional email provider

These environment variables should be configured in Runtime.exs so that it doesn't get set during the docker build stage

Investigate using:
https://github.com/thoughtbot/bamboo vs building out our own adapters

I am leaning towards using bamboo. It's maintained by thoughtbot and we don't have to build our own adapters. 
There are only [4 dependencies](https://hex.pm/packages/bamboo) for bamboo as well so it wouldn't introduce too many 

**Testing**
** NOTE TO CONTRIBUTORS:
This feature is very manual testing heavy since these transactional email are critical for our users. Please include screenshots of the emails being sent when this is working. For both mailgun and SMTP we don't want to break our existing users (DM Kam on Slack for a Mailgun keys for testing) 

Three features we send transactional emails:
1. When a customer receives a message from customer support and they don't see it after 2 minutes an email gets sent to the users see : https://github.com/papercups-io/papercups/blob/master/lib/workers/send_conversation_reply_email.ex
2. When a company needs to invite a customer in the Accounts/Overview page we send an invitation email. See https://github.com/papercups-io/papercups/blob/master/lib/workers/send_user_invitation_email.ex
3. We send password reset emails 

Also need to make sure these manual end to end tests work for both Mailgun and SMTP

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

fullpwemium:
Thank you @cheeseblubber 

I would have a go at this myself but my first exposure to Elixir is Papercups.io! 

Amazon SES is my desired service for this

cheeseblubber:
Ok sounds good most likely bamboo seems like the way to go since it would mean supporting SES with minimal work as well.

[...]

PR #838, opened 2021-05-15
Add an option to use SMTP when delivering emails

fmterrorf:
### Description
Add an option to use SMTP when delivering emails by using the SMTP adapter. 

To use SMTP you need the following environment variable set
`MAILER_ADAPTER` - `Swoosh.Adapters.SMTP`
`SMTP_HOST_ADDR` - (optional) defaults to `mail`
`SMTP_USER_NAME` - required
`SMTP_USER_PWD` - required
`SMTP_HOST_SSL_ENABLED` - (optional) defaults to false
`SMTP_RETRIES` - (optional) defaults to 2
`SMTP_MX_LOOKUPS_ENABLED` - (optional) defaults to true

### Issue
#769 

### Screenshots
Sent using SMTP 
image

Sent using the Mailgun adapter
image

## Checklist

- [x] Ran `mix format`


config/runtime.exs
```diff
@@ -78,7 +78,7 @@ end
 
 # Optional
 sentry_dsn = System.get_env("SENTRY_DSN")
-mailgun_api_key = System.get_env("MAILGUN_API_KEY")
+mailer_adapter = System.get_env("MAILER_ADAPTER", "Swoosh.Adapters.Mailgun")
```
fmterrorf, line 81: Use the Mailgun as a default to make this change backwards compatible. 

I'm thinking if we could use the Local adapter as a default https://hexdocs.pm/swoosh/Swoosh.Adapters.Local.html. That way we dont need a try/catch when calling `deliver` when running locally.

fmterrorf: The nice about the local adapter is that you can add a route when running dev to view the emails that are sent https://hexdocs.pm/swoosh/Plug.Swoosh.MailboxPreview.html

reichert621: @fmterrorf i like this idea! would you mind adding it to this PR, or would you prefer handling in a separate PR?

fmterrorf: I can add it to this PR!

fmterrorf: @reichert621 here's a screenshot of the mailbox preview 
image

config/runtime.exs
```diff
@@ -99,13 +99,33 @@ config :logger, Sentry.LoggerBackend,
...
+      port: System.get_env("SMTP_HOST_PORT", "25"),
+      username: System.get_env("SMTP_USER_NAME"),
+      password: System.get_env("SMTP_USER_PWD"),
+      ssl: System.get_env("SMTP_HOST_SSL_ENABLED") || false,
+      tls: :if_available,
+      retries: System.get_env("SMTP_RETRIES") || 2,
+      no_mx_lookups: System.get_env("SMTP_MX_LOOKUPS_ENABLED") || true
+
+  "Swoosh.Adapters.Local" ->
+    config :swoosh, serve_mailbox: true, preview_port: 4001
```
fmterrorf, line 123: We're going to use another port for the mailbox previewer. I tried using the default 4000 port, but for the routes for previewing the mail does not work, so i decided to put it on a different port. Now we can go to `http://localhost:4001/mailbox` to preview the emails that are sent

lib/chat_api/emails.ex
```diff
@@ -111,11 +111,9 @@ defmodule ChatApi.Emails do
 
   @spec deliver(Email.t()) :: deliver_result()
   def deliver(email) do
-    # Using try catch here because if someone is self hosting and doesn't need the email service it would error out
-    # TODO: Find a better solution besides try catch probably in config.exs setup an empty mailer that doesn't do anything
```
fmterrorf, line 115: Removed the following comments because we will use the swoosh local adapter as a default

config/runtime.exs
reichert621: nice! should we use an env variable for this and default it to 4001?

config/runtime.exs
```diff
@@ -99,13 +99,36 @@ config :logger, Sentry.LoggerBackend,
...
+      password: System.get_env("SMTP_USER_PWD"),
+      ssl: System.get_env("SMTP_HOST_SSL_ENABLED") || false,
+      tls: :if_available,
+      retries: System.get_env("SMTP_RETRIES") || 2,
+      no_mx_lookups: System.get_env("SMTP_MX_LOOKUPS_ENABLED") || true
+
+  "Swoosh.Adapters.Local" ->
+    config :swoosh,
+      serve_mailbox: true,
+      preview_port: System.get_env("MAILBOX_PREVIEW_PORT") || 4001
```
fmterrorf, line 125: cast port to an int

config/runtime.exs
```diff
@@ -99,13 +99,36 @@ config :logger, Sentry.LoggerBackend,
...
+      username: System.get_env("SMTP_USER_NAME"),
+      password: System.get_env("SMTP_USER_PWD"),
+      ssl: System.get_env("SMTP_HOST_SSL_ENABLED") || false,
+      tls: :if_available,
+      retries: System.get_env("SMTP_RETRIES") || 2,
+      no_mx_lookups: System.get_env("SMTP_MX_LOOKUPS_ENABLED") || true
+
+  "Swoosh.Adapters.Local" ->
+    config :swoosh,
+      serve_mailbox: System.get_env("LOCAL_SERVE_MAILBOX", "true") == "true",
```
fmterrorf, line 124: Had to convert this to boolean & the `preview_port` below to int. The local adapter does not accept strings unlike the SMTP adapter

reichert621 (approved):
looks great! 🎉

PR #883, opened 2021-06-16
Adds Getting Started page and renames Account sub-menu to Settings

rhonsby:
### Description

This commit adds a new Getting Started page that acts as a checklist of things users should do to make use of Papercups. In addition, it renames the Account sub-menu to Settings, moves Billing into the Settings sub-menu, and renames Getting Started under Settings to Chat Widget to better reflect what the page it navigates to is.

**Note:** I still need to add the logic to check off each step on the Getting Started page, but I think we can ship this as is while we work on that.

### Issue

https://github.com/papercups-io/papercups/issues/879
https://github.com/papercups-io/papercups/issues/877

### Screenshots

CleanShot 2021-06-16 at 13 44 56@2x

## Checklist

- [ ] Everything passes when running `mix test`
- [ ] Ran `mix format`
- [x] No frontend compilation warnings

Issue #879, opened 2021-06-14
Set up new "Getting started" onboarding UI


reichert621:
**Problem**
Our "Getting started" page just focuses on chat widget settings at the moment, and doesn't really guide users through the full flow of taking advantage of our features (e.g. Slack, email notifications, etc)

**Solution**
Set up a new "Getting started" page, maybe take inspiration from Slack:

![slack-onboarding-ui](https://user-images.githubusercontent.com/5264279/121941185-f3eb6200-cd1c-11eb-83fc-627cc6c15bc3.png)

- This page will consist of the Basic steps and Advanced steps above with each step either checked or unchecked to show that they've completed them. It will make it clearer to the users what they should be doing to make full use of Papercups.
- Consider adding a notification icon on Getting Started to notify user of number steps they need to take (e.g. `Getting started (3/5)`)

**Basic steps we want new users to take:**

1. Configure the chat widget
2. Install the chat widget on their site
3. Invite teammates
4. Configure their personal settings
    - Full name
    - Display name
    - Profile image URL

**Advanced steps we want new users to know about:**

- Set up integrations (e.g. Slack, Gmail, sms, etc).
- Configure StoryTime to view live sessions
- Upgrade their subscription plan for access to more features.

Reasoning: making it a top-level menu will make it clearer there's a page where users can go to in order to set up their usage of Papercups, having it at the top will make it more prominent, and breaking it down into basic and advanced steps with checkmarks will make it easier for users to know what they can and should do.

**Additional context**
https://github.com/papercups-io/papercups/issues/870
https://www.notion.so/6-9-Revamp-Onboarding-6ee0739745814f8ebfa7d17824a15f5d

**Testing**
Please include screenshots in the PR!

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

Issue #877, opened 2021-06-14
"Account" submenu refactoring


reichert621:
**Problem**
The "Account" submenu is a little confusing and hard to navigate. There's lots of room for improvement!

**Solution**
- Rename Account sub-menu to Settings and treat that sub-menu to consist of setting-specific pages only.
- Move new Settings sub-menu to the bottom of the sidebar.
- Rename old Getting Started sub-menu item inside of Settings to Chat widget (or something similar).
- Move "Billing" into this new Settings submenu?

Reasoning: the Account sub-menu is basically where all settings are located. We might as well call it settings and store each setting into its own sub-menu item to further break it down.

**Additional context**
https://github.com/papercups-io/papercups/issues/870

**Testing**
Please include screenshots of the updates in your PR!

**Questions, or need help getting started?**
Feel free to ask below, or [ping us on Slack](https://join.slack.com/t/papercups-io/shared_invite/zt-h0c3fxmd-hZi1Zp8~D61S6GD16aMqmg) :)

(You can also check out our [CONTRIBUTING.md](https://github.com/papercups-io/papercups/blob/master/CONTRIBUTING.md))

PR #884, opened 2021-06-16
Refactors GettingStarted.tsx and adds checking off of the steps

rhonsby:
### Description

In https://github.com/papercups-io/papercups/pull/883, we added the Getting Started page with a checklist of things users should do. This commit adds an endpoint for checking the status of each of those setup steps and uses it to check off each step that's already been completed. In addition, I also refactor `GettingStarted.tsx` to reduce the complexity and size of the `GettingStarted` component.

### Issue

https://github.com/papercups-io/papercups/issues/879
https://github.com/papercups-io/papercups/issues/877

### Screenshots

CleanShot 2021-06-16 at 17 41 27@2x

## Checklist

- [x] Ran `mix format`


lib/chat_api_web/controllers/setup_status_controller.ex
```diff
@@ -0,0 +1,73 @@
...
+        has_integrations: has_integrations?(account_id),
+        installed_chat_widget: installed_chat_widget?(account),
+        invited_teammates: invited_teammates?(account_id),
+        upgraded_subscription: upgraded_subscription?(account)
+      })
+    end
+  end
+
+  @spec configured_profile?(number()) :: boolean()
+  def configured_profile?(user_id) do
```
reichert621, line 35: i'm not always good about doing this myself, but this and the methods below can probably be marked with `defp` (for private)

lib/chat_api_web/controllers/setup_status_controller.ex
```diff
@@ -0,0 +1,73 @@
...
+         %{account_id: account_id} <- current_user do
+      account = Accounts.get_account!(account_id)
+
+      json(conn, %{
+        configured_profile: configured_profile?(current_user.id),
+        configured_storytime: configured_storytime?(account_id),
+        has_integrations: has_integrations?(account_id),
+        installed_chat_widget: installed_chat_widget?(account),
+        invited_teammates: invited_teammates?(account_id),
+        upgraded_subscription: upgraded_subscription?(account)
```
reichert621, line 29: nit - i'm not always good about this myself, but i think it's nice to name boolean fields with `is_`/`has_`/`should_` prefixes to make it really obvious that the fields are boolean, e.g.
```javascript
{
  has_configured_profile: boolean; // or `is_profile_configured`
  has_configured_storytime: boolean; // or `is_storytime_configured`
  has_integrations: boolean;
  is_chat_widget_installed: boolean;
  has_invited_teammates: boolean;
  has_upgraded_subscription: boolean;
};
```
(otherwise i might expect `invited_teammates` to return the list of teammates, or `upgraded_subscription` to return a subscription object, etc.)

test/chat_api_web/controllers/setup_status_controller_test.exs
```diff
@@ -0,0 +1,219 @@
+defmodule ChatApiWeb.SetupStatusControllerTest do
```
reichert621, line 1: hooray for tests! 🎉 🎉 🎉

reichert621 (commented):
nice! super excited for these improvements 🚀 

code looks good 👍 mostly just nitpicky stuff, curious to hear your thoughts @rhonsby

assets/src/api.ts
```diff
@@ -1590,3 +1591,16 @@ export const deletePersonalApiKey = async (
...
+
+export const getSetupStatus = async (
+  token = getAccessToken()
+): Promise<SetupStatus> => {
+  if (!token) {
+    throw new Error('Invalid token!');
+  }
+
+  return request
+    .get(`/api/setup_status`)
```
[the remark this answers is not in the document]
rhonsby: I like it! I can see how `setup` can be ambiguous. Thanks for the suggestion.

lib/chat_api_web/controllers/setup_status_controller.ex
rhonsby: Yeah, I totally agree actually that boolean fields should always be prefixed with one of those prefixes. I also am not always good about it 😅 I'll update it. Thanks for the suggestion.

reichert621 (approved):
:tada: 🚀

Issue #937, opened 2021-08-23
Support email forwarding with Papercups


cheeseblubber:
Allow the user to respond to emails with Papercups

1. Company specifies which emails they want to forward to Papercups support@somecompany.com
2. Papercups generates an unique id for the email to be forwarded probably company-name_some_id@chat.papercups.io 
3. When a customer sends an email it gets forwarded in to Papercups
4. When a user responds to the email it replies back as company-name_some_id@chat.papercups.io  with the support@somecompany.com cc'd

Architecture:
We're going to be using Amazon SES service to receive emails.
Setup:
1. Generate an email id when a company sets up email forwarding
2. Store the mapping between user's support email and generated Papercups Email

Receiving a Messaging:
1. When a customer sends an email it auto saves in S3 with aws SES
2. When it gets saved in S3 a lambda function triggers and hits a webhook to Papercups endpoint with the email payload
3. Papercups reads the email and a new message/conversation/customer gets created

Responding to a message
1. User responds to the message via Papercups 
2. We send an email back via AWS SES api with the message and any attachments. We cc the original email address and the email will be sent from the generated papercups id

PR #948, opened 2021-08-30
SES email forwarding

reichert621:
### Description

Backend APIs and setup for supporting email forwarding with SES

### Issue

https://github.com/papercups-io/papercups/issues/937

### Screenshots
**Setting up forwarding addresses:**
Screen Shot 2021-08-30 at 3 32 06 PM

**Sending and receiving messages in dashboard:**
Screen Shot 2021-08-30 at 3 31 51 PM

## Checklist

- [x] Ran `mix format`


cheeseblubber:
* Forwarding address should be called email under the settings tab like how we have Chat widget. 
* Maybe move forwarding email and chat widget next to each other
* On /settings/forwarding-addresses page it should say "Forward Email" 

Maybe we should add some instructions telling them to forward the address? 
* Like when you press new it generates an address and ask them to forward that email address?

